PR #17, opened 2021-02-12
Possible solution to the overview background actor smooth fadeout problem

yozoon:
Hi! I just played around a little with the code of your extension - I really like the code quality (much better than what I created years ago with Blyr ;) ). I think I found a solution to the problem of overview background actors not properly fading out when we try to animate them. The source of the problem is that during the whole animation in _unshadeBackgrounds, the real desktop background is invisible, as it is a part of the global.window_group that only becomes visible again when the transition from the overview screen completes. So while we fade out the overview backgrounds there is no actual content behind them, therefore we see the graphics garbage (stuff that was in the framebuffer during the previous frame).
So to work around this problem, I move the backgroundGroup one level up, so that it is a sibling instead of a child of the window_group. I also move it behind all other actors at that level to prevent it from covering up the whole screen and making interactions with the desktop impossible. Now when the transition in _unshadeBackgrounds occurs, the actual desktop background is behind the blurred overview backgrounds, thus no graphics garbage is generated.

The limited testing I did showed no problems under XOrg and Wayland, and other shell functionality also doesn't seem to be affected from moving the background.

P.S.: Yesterday I marked the Blyr extension as _unmaintained_ because I really think that for shell versions >= 3.36 it is superseded by your extension, and I don't have the time to manage my codebase anymore. So I also added a link to your extension to the readme of Blyr so that people can find it more easily. :)

Cheers,
Julius

### DIFF
--- a/src/overview.js
+++ b/src/overview.js
@@ -4,6 +4,7 @@ const St = imports.gi.St;
 const Meta = imports.gi.Meta;
 const Shell = imports.gi.Shell;
 const Main = imports.ui.main;
+const Overview = imports.ui.overview;
 const Clutter = imports.gi.Clutter;
 const GLib = imports.gi.GLib;
 const Gio = imports.gi.Gio;
@@ -33,7 +34,6 @@ const old_shadeBackgrounds = Main.overview._shadeBackgrounds;
 const old_unshadeBackgrounds = Main.overview._unshadeBackgrounds;
 
 // numeric values
-const ANIMATION_DURATION = 200;
 let sigma = 30;
 let brightness = 0.6;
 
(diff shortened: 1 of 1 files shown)


aunetx:
Hey!!
Hello, I'm quite excited to see you helping in my extension (I decided to create it after seeing yours, and I never though one day you would see mine)

Thanks a lot for this fix, I'm going to test it very quickly (although the code seems neat, you even fixed my hardcoded animation time :) ) and merge it right now!
I'm really thankful you helped me on this particular problem, I really did not know how to find a clean solution for this... So thanks a lot ^^

And thanks for the attention you have for my extension, but I'm not really sure if mine actually supersedes yours... But maybe one day!

Anyway, I'm gonna test this so thanks again.
Aurélien

PR #26, opened 2021-04-16
Added support for Gnome-shell 40

hmmanshh:
included gnome-shell 40 in metadata


aunetx:
Thanks for the attention to my project! However, this extension is still not compatible with gnome-shell 40, even with these changes... To make it compatible, I would need to find a way to blur the overview, and to restore the blur on the panel after leaving overview :/

Issue #31, opened 2021-04-20
"no artefacts" option makes the extension crash


aunetx:
See https://redd.it/muxj77 for more infos

CorvetteCole:
Oh man I literally had this exact discussion with the gnome devs at one point. Part of why I stopped working on blur provider. Got very complicated and I didn't have time. I think I'd like to migrate the intention of that extension in to this one over a bit. I think being able to arbitrarily provide blur to applications would be a good next step and I won't have to be alone doing it. 

You are the project maintainer, so please let me know if that sounds good to you. If so I'll start working on it!

aunetx:
Yeah, I understand the "got very complicated"... Even just porting to gnome 40 took me almost 2 days, and it's still not finished due to this bug...
But overall I quite like working with gnome-shell in the sense that I use the software I help to extend every day!

If you want to merge your extension into mine, that would be really a great addition! If you need help, don't hesitate to ask me of course :)
Would you like to be granted write permissions to this repository now or maybe later?

For my part, I think I will try (after fully porting to 40) to add some settings (for #30, #22 and #14 notably), and try to fix some compatibility issues (for #10 and #25, and for dash-to-dock, as it is still not stable for gnome 40)

So tell me if you need anything to start working on this!

CorvetteCole:
great to hear! I'll make a separate issue so I can track the tasks I have to do. I never fully finished blur-provider so this will be a good opportunity for me to implement the last couple things I had to do (ability to manually select an application to blur, better Wayland support, better fix for blur jumping in front of actors).

Regarding write-access, I don't need it immediately. I'm happy to work on it in a fork and then submit a pull request for now. I think once I've got my initial implementation integrated in write-access might be nice to have but until then no worries.

I've got to take a fine look at the code in this extension so I can try to match the coding style. Anything special I need to know?

This should be cool!

aunetx:
Thanks a lot for this, Blur my Shell should then be a lot more complete than now :)
I understand for the write access! And tell me if you need to contact me in a more direct way than github issues (eg discord, element, what you need), as you may encounter some hard time with my extension :(

Well, I guess the main thing is to create connections with the `connections.js` helper so they are removed when the actor is destroyed or the extension disabled, like this (once it is added to `this`):

```js
this._connections.connect(actor_to_connect_to, 'signal', () => {
    // what you want
});
```

Also, if possible (to make logging easier), you can add the following method to your classes:

```js
_log(str) {
[...]

Issue #24, opened 2021-03-27
Support for Gnome 40


aunetx:
The extension still does not support Gnome 40:
  - no GTK4 preferences windows
  - no idea how it looks in it, as overview/dash changed

RandomnessRobin:
Hoping to see this extension work in gnome 40 soon. Not sure of that will happen. But i am hoping it does.

aunetx:
I hope too, and of course it will happen!
For the moment, those are the bugs I have:

- [ ] settings need to be ported to GTK4 (should be OK)
- [ ] blur needs to be restored to the top panel after leaving the overview (it is currently not the case in gnome 40)
- [ ] proper animations should be added to the panel
- [ ] overview should be blurred: for this, I need to be able to add an image as a background of the overview (so that I can blur it after, which is easy one the image is there). It would look something [like this](https://redd.it/mtd057)

aunetx:
Panel is done, overview should be OK soon. However, I have a problem with the preferences (I can't open any preference of extension in my testing VM, so I don't know if mine is working...)

aunetx:
Nearly good, it works really better than for gnome 3.38 actually :)
I just need to make sure settings work, and it's good for gnome 40!

To test it, install from the branch "gnome-40"

Saren-Arterius:
Will the new design be like this? https://www.reddit.com/r/gnome/comments/mtd057/in_love_with_this/

aunetx:
Exactly like this yes :)

Saren-Arterius:
That would be holy! Glad to hear! Thanks

Saren-Arterius:
@aunetx I see the panel blur working, but overview's background is still plain color instead of blurred wallpaper. Am I missing something? Or is it gnome theme dependent? Thanks

aunetx:
@Saren-Arterius are you sure that you switched to the "gnome-40" branch?

Saren-Arterius:
Sure

在 2021年4月19日週一 19:37，Aurélien Hamy ***@***.***> 寫道：

aunetx:
If you want to try with this version (you can install it with `gnome-extensions install blur-my-shell@aunetx.zip --force`):
[blur-my-shell@aunetx.zip](https://github.com/aunetx/blur-my-shell/files/6336042/blur-my-shell%40aunetx.zip)

Saren-Arterius:
It is working now. I had to untick and tick overview however. I guess leftover perf caused this.

[...]

PR #34, opened 2021-04-20
Add support for Gnome 40

aunetx:
See #24 for the issue

To this moment, the only big issue that remains is #31, but the extension is stable on gnome 40 if this option is temporarily disabled.


CorvetteCole (approved):
I took a look through these changes, from an extension reviewers perspective everything looks good!

aunetx:
Thanks a lot, do you think it is a good idea to wait for #31 to be resolved before merging this?
I talked to gnome devs yesterday night, but did not find a real solution to it (from what I understood, I would need to extend the `Shell.BlurEffect` actor, and modify its paint vfunc... but that's still some witch language for me, my 'paint' signal workaround was simpler) :)

aunetx:
@CorvetteCole I pushed this to extensions.gnome.org, and the support for dash-to-dock seemed good enough to merge this pull request :)

CorvetteCole:
I approved the extension update! I think I'll rebase my branch on to a new branch called application-blur or something. About the vfunc, I remember being told about that and trying to figure it out but I got stuck in the past. I'm going to take a look at it again, but it seems like some dark voodoo witchcraft to me

aunetx:
Thanks a lot, that's really cool to be able to update the extension so quickly :)
And of course for your branch, do what you need ^^

For the vfunc, I think I kind of understand the principle, but I really don't know how to implement it... For the moment I will add an option to have static blur for the top panel instead of dynamic blur, so there is no more artifact (and it does not change a lot of things, excepted if the use uses something like hide-my-topbar, but that's okay for the moment)

PR #41, opened 2021-04-21
Signals are not properly disconnected, can be duplicated

CorvetteCole:
NOTE: Pull request name and purpose changed to reflect commits that it actually contains

I have allow edits by maintainers on, so you should be able to edit this as desired. I attempted to disable and re-enable the panel blur as a rough fix if the state of dash-to-panel changed (and was enabled).

Ran in to two major issues.

- Blur doesn't get applied to the panel with this disable and enable sequence (I haven't actually been able to get it applied at all).

- The connection to extension changed seems to activate 3+ times when turning dash to panel on for the first time, this is sub-optimal but we could always workaround with a ratelimit if we decide this is the way to do this


aunetx:
A workaround would maybe be to add a delay before disabling/enabling the extension, right? As I guess the extension takes some time to be enabled, so the panel blur might be changed before the panel is replaced

Maybe something like:

```js
setTimout(() => {
    this.disable();
    this.enable();
}, 500);
```

With maybe some other value than 500, bigger if it helps or less if possible

CorvetteCole:
well I'll be darned, isn't that interesting
Screenshot from 2021-04-21 22-57-42

Further testing reveals that it... works perfectly?!!?? No seriously, no matter what order etc this works now. Very very neat! Only one problem, but this is probably a pretty straightforward fix. There is no blur on the panel on the secondary monitor(s).
Screenshot from 2021-04-21 22-59-16

CorvetteCole:
Although I'm not sure the timeout is working as intended. See log message: 
```
JS ERROR: Exception in callback for signal: extension-state-changed: ReferenceError: setTimeout is not defined
                                             enable/<@/home/coleg/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/panel.js:49:17
                                             _emit@resource:///org/gnome/gjs/modules/core/_signals.js:114:47
                                             _callExtensionEnable@resource:///org/gnome/shell/ui/extensionSystem.js:170:18
                                             _onEnabledExtensionsChanged/<@resource:///org/gnome/shell/ui/extensionSystem.js:493:35
                                             _onEnabledExtensionsChanged@resource:///org/gnome/shell/ui/extensionSystem.js:493:14
                                             createCheckedMethod/<@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:516:46
                                             enableExtension@resource:///org/gnome/shell/ui/extensionSystem.js:194:29
                                             EnableExtension@resource:///org/gnome/shell/ui/shellDBus.js:310:38
                                             _handleMethodCall@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:310:38
                                             _wrapJSObject/<@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:387:34

```

CorvetteCole:
gjs doesn't have timeout. Luckily, there is an implementation of it I found that we can likely use instead.
https://gist.github.com/andyholmes/582c29facbdbe67048c831a6370173eb

Will test this tomorrow!

CorvetteCole:
I actually just found that you used this previously here: https://github.com/aunetx/blur-my-shell/blob/f48d7bcdac060d31ece008eb2d8eff90ec256819/src/overview.js#L12

If it sounds good to you, I think I may split off this function in to a generic utilities.js class that we can just import if needed.

CorvetteCole:
I changed the name of this pull request to reflect the work I did... on the wrong branch. Dash to panel work is still happening but this feels important to get merged (it actually has a performance impact).

CorvetteCole:
I butchered the rebase once so I have some duplicate commits but everything should be good for a merge (I triple checked)

CorvetteCole:
Please make sure to review this as it makes some changes to the extension structure which you will probably want to be familiar with. This does not change how we handle connecting to signals inside of modules so we will not have to code anything differently there

CorvetteCole:
The PR also contains files for autocomplete in the @types folder. These are auto-generated, do not affect the code, and probably don't need to be looked at. Will eventually put some instructions in the README to give potential other developers super easy setup instructions

aunetx (approved):
This is really very very great :)
Maybe we would need remove the log in extensions init, as we need not to do anything at all there, right?
And I will maybe later add the @types support for vscode too (I intended to do so at one point, so this is great that you added this)

aunetx:
Would you like to merge right now? I think this should be really neat, but I don't know if you plan to improve even more this PR

aunetx:
By the way, could you give me (if you have) the link where you found the informations for adding typescript's @types?
I don't find how to properly add it for vscode (which would be helpful for me as I develop with it)

CorvetteCole:
Sure, I'll try and make some instructions for using the @types. I couldn't find any so I kind of just figured it out myself. Not sure I did it right but I am going to add it to the README. By the way for reference this website describes what we should and shouldn't do in init() https://wiki.gnome.org/Projects/GnomeShell/Extensions/Review.

I am going to remove the init log and then merge this PR

aunetx:
@CorvetteCole I pushed it to gnome extensions, sorry for the latency...
I added @types support on vscode (was actually simpler than expected), it's nice!

By the way, do you think it has any interest to type-check the code (by using typescript for example, or something like this)?
I don't believe that the extension is complicated enough to need it, but this may be good... I don't really have a strong opinion

CorvetteCole:
sorry I took so long on the howto for the @types support! I still intend to write up instructions on how to generate those definitions for the future as the libraries upgrade :)

CorvetteCole:
On the Typescript front... I love strict code, but it feels like a lot of extra work to do that at the moment. I say we just be careful for now

aunetx:
I removed then @types files from being tracked in git, because it looked really a lot (more than 300000 loc to download when cloning), so for the moment I prefer letting them untracked.

We might change this later (actually, especially when I get to create the correct config files for vscode, because for the moment it makes every method names in camelcase, and gnome-shell uses snake case so there is really little use).

PR #39, opened 2021-04-21
Merge functionality of blur-providers extension in to this one

CorvetteCole:
This draft PR will track the progress of work to port the existing application blurring functionality of the [blur-provider](https://github.com/CorvetteCole/blur-provider) extension to blur-my-shell. 

Beyond simply porting existing functionality, several other changes need to be done. These have been summarized below:

- [x] Port existing application blurring on X flag functionality
- [x] Improve Wayland support (https://github.com/CorvetteCole/blur-provider/issues/2) (needs testing!)
  - [x] Figure out a way for applications to request blur beyond X flags. (e.g. an application not running within XWayland or X
- [ ] Add manual user application selection for blurring (https://github.com/CorvetteCole/blur-provider/issues/3)
  - [ ]  Add application selection to the extension settings
- [ ] Keep blur behind target actors in overview (right now it just kind of disappears)
- [x] ~After switching to a workspace and back, blur appears on top of the target actor until the screen is interacted with~ Fixed In Gnome 40
- [x] ~When switching focus rapidly between windows, blur can blink in front of all windows briefly~ Fixed In Gnome 40

PR #52, opened 2021-04-25
Add effect to emit 'paint' signal

### LINKED ISSUE #31: "no artefacts" option makes the extension crash

aunetx:
I successfully added the effect to fix #31, and it works great!
I tested it for the panel, it works as great as the ancient 'paint' signal (so there are still some artefacts, due to windows shadows...).

We may need a way to instantly add this effect and connect it, keep track of it and maybe clean it afterwards? Just like we do for the connected signals, it would simplify the design a bit.

### DIFF
--- a/src/connections.js
+++ b/src/connections.js
@@ -39,6 +39,25 @@ var Connections = class Connections {
         this.process_connection(actor, id);
     }
 
+    disconnect_all_for(actor) {
+        actor_connections = this.buffer.filter((infos) => {
+            infos.actor == actor
+        });
+
+        actor_connections.forEach((connection) => {
+            // disconnect
+            try {
+                connection.actor.disconnect(connection.id)
+            } catch (e) {
+                this._log(`error removing connection: ${e}; continuing`)
+            }
+
+            // remove from buffer
+            let index = this.buffer.indexOf(connection);
+            this.buffer.splice(index, 1);
+        })
+    }
+
     disconnect_all() {
         this.buffer.forEach((connection) => {
             try {
--- a/src/dash_to_dock.js
+++ b/src/dash_to_dock.js
@@ -7,6 +7,7 @@ const Signals = imports.signals;
 const Me = imports.misc.extensionUtils.getCurrentExtension();
 const Settings = Me.imports.settings;
 const Utils = Me.imports.utilities;
+const PaintSignals = Me.imports.paint_signals;
 let prefs = new Settings.Prefs;
 
 const default_sigma = 30;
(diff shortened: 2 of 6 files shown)


CorvetteCole:
I think I understand what you mean, but it looks like you are already adding the connection to that modules connection object. What more did you have in mind?

aunetx:
What I was thinking about is simply that, when we remove the blur effect from the actor, with this method we will need to remove the `EmitPaintSignal` effect too (or it could be duplicated, or cause errors/overhead...)

So I was thinking about a method very similar to the one in connections.js, but maybe adapted to use this thing?

CorvetteCole:
oh that's an interesting point, we don't want to needlessly emit extra signals. I'm not sure how we would implement this, but I think it is definitely a good idea. Maybe a PaintSignals.js class. Did you want to code that or should I put it on my to-do?

aunetx:
Yeah maybe a dedicated class... Will you use this effect a lot for #39? If yes, this should definitively be added, but if not maybe just keeping track and removing the effects within the extension (for panel and overview) could be enough...

Anyway, if you want to work on this you can do it! And if not I will do this in some days/weeks anyway :)

CorvetteCole:
It is pretty essential for #39. Artifacts on the blur effect of windows are really common since they are often over other actor shadows and moved around a lot. So will definitely be used there! I'm not going to start working on the class yet since I am pretty busy at the moment, but I will let you know before I do if I get to it first so we don't duplicate efforts :)

aunetx:
I'll start to see what I can do about the paint signal :)

This is a good idea, it may be hard to keep track of everything else! Would you like to try to do this?

aunetx:
I think this is good enough to be merged! But I'll wait for your review first, don't want to merge it if the model is not adapted to #39 :)

By the way, if you have dash-to-dock, you can test it with this PR, I did not test it enough because I don't have it installed... But it should be ok though!

aunetx:
I will try to entirely remove artefacts BTW, I think that I have a very hacky solution (being to just track windows, and update the panel blur when they are moved... because their shadow is still creating artefacts, so I may be better with this).

CorvetteCole:
sorry for such a late response. I  started a new internship only a few days ago and have been very very busy with it and moving. I'm going to be spending my weekends working on this again after the next two where I will be away. sorry again

CorvetteCole (approved):
I like these changes and think this is a pretty good way of handling things. A great workaround

aunetx:
Cool thanks for the review, will merge this right now :)

And don't worry I really understand for the delay! I'm not very focused on the extension too, feel free not to answer for as long as you want :)

PR #63, opened 2021-04-29
Application blur (With fixed wayland!)

ckissane:
+ Settings work
+ Toggling works
+ Changing blur brightness and sigma works

does all applications by default
Screenshot from 2021-04-29 14-59-20


saaditory:
Awesome, I appreciate the work behind this.

Can you tell me which shell theme you are using?
Like It's not working for me:
Screenshot from 2021-04-30 03-19-30
Also it's a complete rectangle, screenshot is making half of it transparent.

I'm using GNOME 40.

CorvetteCole:
@saaditory this is a bug I experienced with those running Wayland. It is part of the reason that I am taking the time to rewrite the functionality of the blur-provider extension before merging it in #39

CorvetteCole:
@ckissane This is fun, but I am still working on rewriting pretty much the whole thing in #39 , to fix all the previous problems it had. I'm not sure merging this would be the right move in the long run with that being worked on

ckissane:
@saaditory  use https://github.com/owozsh/Myopia-Theme transparent (T) version

CorvetteCole:
If you'd be okay with it, I would love to use bits and pieces from what you've written here to simplify portions of my efforts in the other PR. For now I am not going to merge this and continue working on my PR, but if @aunetx feels different they are the maintainer

ckissane:
@saaditory any luck with the theme?

saaditory:
I personally don't think that this PR is anywhere close to being usable, even on Xorg.

Xorg test:
1. Works without `application blur` option (just sigma is 0).
2. Enabling of fractional scaling doesn't work, (Crashes settings app).
3. Random visual glitches can appear while enabling & disabling `application blur` option.
4. ? Screenshot from 2021-04-30 04-46-43

ckissane:
@saaditory Could you explain what you mean by fractional scaling so that it may be fixed?

IFAIK fractional scaling is experimental in gnome and integer scaling (100%,200%,300%) is working fine for me.

ckissane:
@saaditory could you also decribe your hardware a bit more? That may be helpful.

ckissane:
@saaditory also if you are wondering why some things are not themed it's because the myopia theme doesn't have gtk-4.0 support but you can spoof it by copying the gtk-3.0 folder inside and renaming it to be gtk-4.0.

You also will have to restart the shell for the theme to apply to the extensions menu I think.

Screenshot from 2021-04-29 17-43-47

ckissane:
@CorvetteCole I fixed wayland

CorvetteCole:
love the work here. I'm trying to figure out how I can merge these efforts with #39 in the quest to get a proper stable version of application blurring in

CorvetteCole:
Also it appears that you are blurring behind ALL applications, which was never the intention of the blur-provider extension, but I think we will want to provide as an option ultimately

ckissane:
true dat

PR #72, opened 2021-05-12
Added gnome 40.1 support

FedShat:
Works fine in Fedora 34, gnome 40.1, so why not bump shell version?
Screenshot from 2021-05-12 15-16-13

### DIFF
--- a/src/metadata.json
+++ b/src/metadata.json
@@ -2,7 +2,8 @@
     "description": "Adds a blur look to different parts of the GNOME Shell, including the top panel, dash and overview.\n\nContains some bugs due to the implementation of the blur effect on gnome shell, see https: //gitlab.gnome.org/GNOME/gnome-shell/-/issues/2857 for more informations.\n\nAdded support for Gnome 40, and a way to entirely remove artifacts from top panel :)",
     "name": "Blur my Shell",
     "shell-version": [
-        "40"
+        "40",
+        "40.1"
     ],
     "url": "https://github.com/aunetx/gnome-shell-extension-blur-my-shell",
     "uuid": "blur-my-shell@aunetx"


CorvetteCole:
version "40" should work on any point release of 40 including 40.1, 40.rc.1, etc. we don't need to specify the point release cause that'll create a cluster of weird dependencies for people running an older version (they won't get updates)

FedShat:
It doesn't, extensions.gnome.org shows error on installing, manual putting into the directory doesn't work too. With this patch works fine.

CorvetteCole:
well that isn't expected behavior, let me ask around real quick to make sure that I'm not remembering things wrong. I'll re-open for now to keep track

aunetx:
Well, I think that we should merge it if it causes an issue... It should at least not introduce any! Thanks ^^

FedShat:
Thank you :)

CorvetteCole:
@FedShat I couldn't reproduce this issue on the latest gnome 40.1 on arch linux... it is weird you had it.

FedShat:
> @FedShat I couldn't reproduce this issue on the latest gnome 40.1 on arch linux... it is weird you had it.
I can't send screenrecord here, would send just screen of an attempt of installing from extensions.gnome.org
Screenshot from 2021-05-13 18-16-22

CorvetteCole:
I have a sneaking suspicion this was about with something else and the shell restart after changing the extension was what "fixed" the issue rather than the change itself. Gnome can be buggy regarding extensions

CorvetteCole:
@aunetx I personally think we should revert this merge because it shouldn't be an issue (and isn't reproducible). Using 40 only is best practice because that implicitly supports all point releases (there is no actual difference in how 40 and 40.1 is handled in Gnome as long as the shell is at least 40.x)

FedShat:
Hm, may be, I will try now manual installing without change, wait

FedShat:
Sorry for bothering, you are right, restart is a thing that fixed.

CorvetteCole:
No worries, always better to know!

PR #76, opened 2021-05-14
Revert "Added gnome 40.1 support"

CorvetteCole:
Reverts aunetx/blur-my-shell#72, we don't need this after all


CorvetteCole:
See the tail end of #72 for more information. This was not an actual problem

Issue #47, opened 2021-04-23
Static panel blur does not take the whole screen with multiple monitors


Caliel666:
![image](https://user-images.githubusercontent.com/38433434/115811598-621f4280-a3c6-11eb-9312-92ffeb4ff3a4.png)

works fine when i switch to a 1080p resolution or use dynamic panel blur, but it's really glitchy even with artifacts hacks on

aunetx:
Could you give me the result of typing `Main.panel.width` in looking glass (`alt+f2`, type `lg`)?
And the result of `Main.layoutManager.primaryMonitor.width` too, so I make sure I change the right thing

aunetx:
Last commits should have fixed this, but if not then please tell it to me :)

Caliel666:
hello, here's the outputs

![image](https://user-images.githubusercontent.com/38433434/115867000-e0a0d200-a410-11eb-94a9-5ae0c1964192.png)

and last commit didn't do anything, i even tried to change some things in panel.js (140-151) to the exact width and height of my resolution and nothing happened

aunetx:
Oh ok, that's a much worse bug than I though... Does your wallpaper resolution match (or exceed) your monitor resolution?
It looks like you are using a 1080p wallpaper, and so when cloning it, the extension does not extend it so it matches the right resolution

aunetx:
I mean, I think, because I can't reproduce this on my machine...

aunetx:
In looking glass, could you tell me the result of those commands:

```js
Main.panel.get_parent().get_child_at_index(0).width
Main.panel.get_parent().get_child_at_index(0).get_child_at_index(0).width
```

And maybe try these commands, to see if it changes anything:

```js
Main.panel.get_parent().get_child_at_index(0).set_x_expand(true)
Main.panel.get_parent().get_child_at_index(0).get_child_at_index(0).set_x_expand(true)
```

Caliel666:
my wallpaper is waaay bigger then my resolution, but even when changing it to a different one doesn't make the panel look correct

outputs:
![image](https://user-images.githubusercontent.com/38433434/115872130-d2a27f80-a417-11eb-9ff4-7855cca10d3a.png)

i think the issue is `Main.panel.get_parent().get_child_at_index(0).get_child_at_index(0).set_x_expand(true)`

Caliel666:
quick update

[...]

PR #82, opened 2021-05-26
Fix static blur being broken in setups in which the main monitor is not the leftmost-topmost one.

DarkRyu550:
This is a fix for the issue I reported in #47.

This bit of code originally set the clip rectangle of the actor responsible for static panel blur to be offset to be the same as the offset of the monitor from the top left corner of the screen. This would break static blur in setups in which the main monitor is not the topmost-leftmost one.

The offsets handed out to the `Meta.BackgroundActor.set_clip` method, inherited from `Clutter.Actor.set_clip`, are specified in [the documentation](https://developer.gnome.org/cluttermm/unstable/classClutter_1_1Actor.html#a04bd53cedd2e186a0e56d93aa6f6ca73) as being offsets from the top-left corner of the _actor itself_. Thus, since the actor responsible for the blur is already positioned at the top-left corner of the main monitor, this offset should be `0, 0`, rather than the position of the actor.

### DIFF
--- a/src/panel.js
+++ b/src/panel.js
@@ -151,8 +151,8 @@ var PanelBlur = class PanelBlur {
         this.background.height = Main.panel.height;
         if (is_static) {
             this.background.set_clip(
-                this.monitor.x,
-                this.monitor.y,
+                0,
+                0,
                 this.monitor.width,
                 Main.panel.height
             );


aunetx:
Oh yeah, I was dumb... Thanks a lot for this fix, it looks perfect!

I just test it on my setup and merge it right after :)

Issue #85, opened 2021-05-30
Panel background doesn't update


shemgp:
Using version 13 here:

![image](https://user-images.githubusercontent.com/1734126/120100803-50734c80-c175-11eb-9560-6e1e8360f4d2.png)

shemgp:
If I restart gnome-shell (alt-f2 r) then it updates. However, after a lock, it gets stuck at the background of the lock. The bug is just with the top panel and not with dash to dock.

KristianUSN:
Same issue, reverting to version 12 fixes it however

aunetx:
Sorry for not answering before... Looks like you found a way to solve this, will merge it soon!

PR #86, opened 2021-06-05
Remove background_panel first to avoid duplicates

shemgp:
Looked at lg why the panel background is not being updated after a lock and noticed that for each lock, the background_panel gets created as a new instance. That means that the old background_panel is still there so that's why the panel background wasn't get updated. 

Here's a workaround which removes the old background_panel(s) before inserting a new one.

Fixes: #85

### DIFF
--- a/src/panel.js
+++ b/src/panel.js
@@ -26,6 +26,7 @@ var PanelBlur = class PanelBlur {
             mode: prefs.STATIC_BLUR.get() ? 0 : 1
         });
         this.background_parent = new St.Widget({
+            name: 'topbar-blurred-background-parent',
             style_class: 'topbar-blurred-background-parent',
             x: this.monitor.x,
             y: this.monitor.y,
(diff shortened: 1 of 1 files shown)


shemgp:
Added commit where dash to dock height is subtracted by panel height instead of only 8. Helps remove dash-to-dock not reaching to top when hide top panel is enabled:
image

aunetx:
Cool thanks, looks great! But does this configuration works also when using dash-to-dock in horizontal mode? Because I originally removed 8px to make it follow the original dock look (horizontal), and idk how it works with this...

If you can test it, I will merge this right after or else I will test it myself (but I don't have the time right now sorry :/ )

shemgp:
Reverting my last commit so that this PR only solves the top panel not getting updated. The other fix will need more work since it didn't work with horizontal mode.

aunetx:
Thanks a lot, this is perfect now :)

PR #87, opened 2021-06-06
Set dash to dock blurred background position depending on its position

shemgp:
I'm using [dash to dock for gnome 40 PR](https://github.com/micheleg/dash-to-dock/pull/1402) and this fixes the wrong position of the dash to dock blurred background for all positions (note that that PR has a bug where the dash has a space on top when position is top).

Before this fix the left position would have like this:
image
The bottom would have a space too:
image
The right would have a space on top too:
image


aunetx:
This looks... Perfect :)
Thanks a lot for this!

Issue #151, opened 2021-11-16
Feature Request: blur the gap between workspaces


K4rakara:
Theres a tiny gap between workspaces that is most visible if you slide between workspaces using a trackpad (slide 3 fingers to the left or right), and atm it is just the default grey that the overview page has.

Heres a screenshot of what I mean:

![image](https://user-images.githubusercontent.com/40474474/141883713-ef69c0d5-3a89-4483-8db2-0c766fbb9d3a.png)

Schneegans:
Thank you for this awesome extension! The next version of the [Desktop Cube Extension](https://extensions.gnome.org/extension/4648/desktop-cube/) will "cubify" this transition between workspaces. The "tiny gap" will become much bigger in this case:

![Screenshot from 2021-12-27 20-28-23](https://user-images.githubusercontent.com/829942/147501630-b17ef406-1876-4e17-a977-ff77d146312e.png)
 
It would be totally awesome to have the blurred overview background in this case!

aunetx:
Sorry for the lack of answers on this repository, I'm honestly submerged by work (preparing for my entrance exams...) :/

I will try to give a look at all the very numerous issues, but I will have a hard time to do all of this alone for the next months to be honest!

-----

@Schneegans, I love (really love) your extensions, thanks for bringing joy to gnome :)

If you have any idea how I can add a background image to the gap (or at least interact with it, I can't even find it in gjs...), that would be super helpful! Else, I will need to do some very big hacks to write/load css files automatically (as I can still interact with it with css, with `.workspace-animations` IIRC, except if you know how I can add css rules in a more simple way). That would be hard and probably unstable, but totally cool and probably make this the first self-modifying extension!!!

Btw, you seem to know a lot about shaders and gjs, do you know if there could be a way to apply a shader (or better, a gnome/mutter effect) on an image and save it on the disk, directly from the extension? This would be the condition sine qua none to use the self-modifying css :)

Sorry for asking all of this, I also write this here to remember to ask on matrix/IRC later if you don't know!

-----

This is a duplicate of #56 btw, but with the current number of issues... I may as well let this open and try to make it to 100 opened issues ! :sweat_smile::tada:

K4rakara:
Don't sweat it! I've been struggling to keep up with my own projects for the same reason, stay strong!

Schneegans:
> Sorry for the lack of answers on this repository, I'm honestly submerged by work (preparing for my entrance exams...) :/

[...]

Issue #56, opened 2021-04-27
Feature Request: blur divison of workspaces


Dziban-dev:
![2021-04-27_08-54](https://user-images.githubusercontent.com/36001673/116263009-41dce400-a736-11eb-9cca-bcf2ebcaffef.png)

it would be be neat to show the blur background when sliding through workspaces,

CorvetteCole:
If you have the static blur checkbox turned off, this should do what you desire. Can you confirm that?

saaditory:
> If you have the static blur checkbox turned off, this should do what you desire. Can you confirm that?

`Static panel blur` just moves the blurred version of the gray area through panel along with workspaces. But the request here is that to **make that separator(gray area) between workspaces transparent/blur**.

CorvetteCole:
Wow, I see what you mean now. I hardly even see that. Well maybe @aunetx would have some ideas about this. I have no idea what part of the shell that is

aunetx:
I really don't know how to change this, but will try to find this :)

However, I don't really know if it would be a good idea to do this... I mean, what should we blur behind it? Should it be a blurred version of the wallpaper, but then which part? It we are changing workspace, then the wallpaper is not static, and I don't know how it would visually look like...

CorvetteCole:
I personally think this should be a very low priority since it is just a gray part. Also not sure what would put there as well

Dziban-dev:
> I really don't know how to change this, but will try to find this :)
> 
> However, I don't really know if it would be a good idea to do this... I mean, what should we blur behind it? Should it be a blurred version of the wallpaper, but then which part? It we are changing workspace, then the wallpaper is not static, and I don't know how it would visually look like...
Well yeah i figured it would be complecated, my idea is just being able to see the blurbackground , so it would look like paralax when workspaces move over it. i mean the blur background would be static , and workpaces are just moving over it. kind of like they do in the overview but fullscreen.

Dziban-dev:
> I personally think this should be a very low priority since it is just a gray part. Also not sure what would put there as well
yeah probably is not even possible to do.

CorvetteCole:
@Dziban-dev did you delete a comment here? I can't see it posted but got an email regarding it. Not a terrible idea, parallax

tommytran732:
> I really don't know how to change this, but will try to find this :)
> 
[...]

PR #148, opened 2021-11-09
Try to fix #137

aunetx:
May fix #137

PR #166, opened 2022-01-07
Add blurred background to gap between workspaces

Schneegans:
I tried the approach mentioned in #151 and it seems to work quite nicely! I tested this on GNOME 40 & 41 and did not observe any issues.

I added the code to `overview.js` so that it gets toggled together with the overview blur. I did this because I think that most users will think that the gray color seen through this gap is basically the overview's background. Alternatively we could introduce a new option, I guess.

What do you think?

Edit: This would also fix the duplicate issue #56.

### DIFF
--- a/src/overview.js
+++ b/src/overview.js
@@ -2,6 +2,7 @@
 
 const { Shell, Gio, Meta } = imports.gi;
 const Main = imports.ui.main;
+const WorkspaceAnimationController = imports.ui.workspaceAnimation.WorkspaceAnimationController;
 
 const Me = imports.misc.extensionUtils.getCurrentExtension();
 const Settings = Me.imports.settings;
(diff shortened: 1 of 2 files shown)


aunetx:
Wow, thanks a lot!! You really are some kind of magician to do this as cleanly as that!

I just pushed a little commit to rename `_insert_child_at_index` (and move it before less important methods), and I made sure that the workspace animation was actually transparent even when Just Perfection forces it to be black :)

I think adding it here is actually the best thing to do, and a switch could be considered -- but maybe rewriting the preferences entirely would be better then, so that there is an actual separation between each component! So for the moment, I guess we can let this like this; I don't think anyone will be annoyed by the addition :)

Thanks a lot once again, this is really super helpful! I will push this to ego very soon, probably along with #163 and #148! This should bring some changes to this extension, and really polish the experience :)

PR #169, opened 2022-01-13
Check workspace setting before blurring workspace switch

### LINKED ISSUE #168: [Bug] Background is blurred when switching between desktops on multi-monitor setups
Since the last update (25 → 26), when switching between desktops on a multi-monitor setup, the background image gets blurred:

https://user-images.githubusercontent.com/28832235/148620964-1e5d5aa3-7b17-40f9-9d27-a1443b5a572f.mp4

Before the update, the background image did not get blurred when switching between desktops without opening the overview.

CodedOre:
This PR adds an check before applying blurring to the workspace switch.

If "workspaces only on primary display" is activated, it will only apply the blur to the primary display, otherwise it behaves as previously.

Fixes #168

### DIFF
--- a/src/overview.js
+++ b/src/overview.js
@@ -55,10 +55,7 @@ var OverviewBlur = class OverviewBlur {
             outerThis._origPrepareSwitch.apply(this, params);
 
             Main.layoutManager.monitors.forEach(monitor => {
-              var blur_this_monitor = true;
-              if (Meta.prefs_get_workspaces_only_on_primary()) {
-                var blur_this_monitor = monitor == Main.layoutManager.primaryMonitor;
-              }
+              let blur_this_monitor = Meta.prefs_get_workspaces_only_on_primary() && (monitor == Main.layoutManager.primaryMonitor);
               if (blur_this_monitor) {
                 const bg_actor = outerThis.create_background_actor(monitor);
                 Main.uiGroup.insert_child_above(bg_actor, global.window_group);
--- a/src/overview.js
+++ b/src/overview.js
@@ -55,11 +55,17 @@ var OverviewBlur = class OverviewBlur {
             outerThis._origPrepareSwitch.apply(this, params);
 
             Main.layoutManager.monitors.forEach(monitor => {
+              var blur_this_monitor = true;
+              if (Meta.prefs_get_workspaces_only_on_primary()) {
+                var blur_this_monitor = monitor == Main.layoutManager.primaryMonitor;
+              }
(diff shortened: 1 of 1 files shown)


aunetx:
Thanks a lot for this fix! I just removed an if to use a boolean test directly, but it looks perfect :)

I merge this now, and will try to push to extensions.gnome.org today!

CodedOre:
> I just removed an if to use a boolean test directly

Actually, it was there for a reason.

If you have workspaces set on all monitors, with your commit it results in the effect being not applied at all:
broken-switch

I have thought about making the check in one line, but using `Meta.prefs_get_workspaces_only_on_primary() && (monitor == Main.layoutManager.primaryMonitor)` would result in following options:

| Setting / Monitor              | Primary (`true`)   | Secondary (`false`) |
| ------------------------------ | ------------------ | ------------------ |
| Workspaces on primary (`true`) | blur monitor       | don't blur monitor |
| Workspaces on all     (`false`) | don't blur monitor | don't blur monitor |

aunetx:
Uh, sorry for that, I really did not think it would have any side effect... I fixed this in fd784f2705cf2f0669a4df611337e713b7807159, and I will maybe find another way to do it later!

CodedOre:
Just came to my mind, but if you want to set the variable in one line, this might work:

```javascript
let blur_monitor = ! Meta.prefs_get_workspaces_only_on_primary() && (monitor != Main.layoutManager.primaryMonitor)
```

By checking if the monitor is not a secondary monitor, we get these possibilities:

| Setting / Monitor      | Primary (`false`) | Secondary (`true`) |
| ---------------------- | ----------------- | ------------------ |
| One Monitor (`true`)   | `false`           | `true`             |
| All Monitors (`false`) | `false`           | `false`            |

If we then negate the result we should get what we want (always blur unless one monitor setting and secondary monitor).

I'll try this quick...

CodedOre:
> Just came to my mind, but if you want to set the variable in one line, this might work:
> 
> ```js
> let blur_monitor = ! (Meta.prefs_get_workspaces_only_on_primary() && (monitor != Main.layoutManager.primaryMonitor))
> ```

Made some test, and it seems to work with this.

So, if you want to reduce the if-clauses to one, I can pull up an PR with this change!

aunetx:
Well, thanks a lot!
If you want to do a PR, that would be perfect! Else I will just commit this when I have my computer with me :)

Thanks a lot for the fix, I really did not have anything to test a multi-monitor setup with me tbh!

PR #170, opened 2022-01-13
Add static blur for dash-to-dock

### LINKED ISSUE #167: Add static blur for dash-to-dock
Now that Dash to Dock seems stable for gnome 40/41, I can provide an option to use a static blur -- just like for the panel.

This would be cool because:
- no more artefacts
- the dash content should be more readable (due to the background being less uniform with the window behind)
- rounded corners!!!

I will however add a switch (just like for the panel), because:
- no more dynamic blur, which was kind of the point of the extension at the beginning
- we're an extension, so I don't have any preferences guideline! :p

The switch could be switched on by default, because it provides a more polished user experience :)

aunetx:
Still does not work well, I can't get to reliably get the final position/size of the dock after startup.

Fixes #167, and maybe a lot of other complaints :)


RyzenDew:
any update on this PR?

ludg1e:
bump, any update?

aunetx:
Superseeded by #533

PR #171, opened 2022-01-13
Blur multiple panels when using Dash to Panel

### LINKED ISSUE #152: Blur multiple panels (Dash to panel)
I have 3 monitors and only the primary monitors panel gets blurred. Would it be possible to blur all of the panels.

aunetx:
Would fix #152, this draft is not working (a lot of things should be changed to get it working), but the main idea is here!


aunetx:
Would fix #295 too

PR #172, opened 2022-01-13
Determine workspace switch blur in one line

CodedOre:
Follow-up to #169, checks if the monitor should be blurred in one line.


CodedOre:
One note to the end:
While testing this solution a bit more, I found that sometimes the blur is not applied on the secondary monitor while having workspaces on all monitors, but that seems not be coming from this code part, but rather the bug you're already noted on your Readme:

> overview blur is transparent on second monitor when using Wayland, sometimes :(

Thought this might be worth mentioning, even though it didn't happened that often to me.

PR #165, opened 2022-01-07
Add dynamic blur

aunetx:
This will fix #90.

PR #180, opened 2022-01-25
Rewrite the preferences to a more modern look, and add options

### LINKED ISSUE #54: Add option to adjust panel brightness/sigma independently
With the way it is now, the text from my panel blends into the background. It would be nice to be able to change the panel settings independently from the other options to be able to alleviate this.

aunetx:
This is a big rewrite of the extension, nearly every component is changed.

This will fix #54, and will permit to complete #39 to select which window is blurred from the preferences :)


aunetx:
@CorvetteCole you may want to check here, a lot have changed, including but not limited to:

- use gresources to load external files (not very useful for the moment, but maybe more when preferences are correctly done), and better makefile
- per-component brightness/sigma preferences if the user wants to (through gsettings for the moment)
- a lot of refactor (limit of 80 characters per line when possible, enforce the use of semicolon everywhere)
- applications blur enhanced:
  - xprop format changed to have a more extensible API (you can see in details in the code, I commented a lot), now you need to set mutter hint to `blur-provider=sigma:20,brightness:0.6` for example, or `blur-provider=s:20,b:0.6` equivalently
  - possibility to select permanently-blurred applications (through gsettings for the moment)
  - fixed some bugs that did not appeared with non-permanently blurred apps

These are some very big changes, including to the PR you submitted, so tell me what you think about all of this!

For the moment, the main part of this PR is not done (you can check preferences... it's pretty empty :p); but all these changes will eventually permit this.

You can test this, it works pretty well (apart from some bugs, like applications blur not being correctly applied after extension restart due to how I connected everything together; I will need to change some more things in order to prevent that).

CorvetteCole:
that's great! I'll definitely take a look. I previously took the sigma from the applications and the brightness from the extension settings so this works for sure.

I'd love to help with the settings rewrite too! Glad to be working with you on this

bertin0:
How would I go about testing the application-blur with this new PR? What else do I need to do other than setting applications-blur to true and the whitelist to my app through gsettings be enough? Also, can I do that in the schemadir `~/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/schemas`, or is it necessary to do it before installing in the build folder?

aunetx:
Well, I guess after installing from this branch, these two things should be enough! You can find your app name with xprop (it's the WM class name iirc), and simply put that with gsettings.

If that works from the extensions dir then that's enough, although I don't remember if I included everything needed in it (but else, just `make build` and take it from `build/schemas`)

bertin0:
Thanks! It does work from the extension folder, my only issue was I didn't have the correct app name, and yes the `WM_CLASS` is the right one. One issue though: doesn't seem to work on wayland. Is it because xprop is X11 specific? If so, is there any way you could implement it on wayland?

aunetx:
In what way doesn't it work in wayland? To get the window name with xprop? Once the settings is updated, there will be an option to directly select the window from there -- so no need to use xprop or anything :)

aunetx:
(and to get the WM name from gjs, I already did that, so the only thing missing is a way to select a window from the prefs, similarly to the window selector for the new screenshot tool)

bertin0:
Blurring alacritty for example doesn't work for me on wayland. I have set the whitelist to Alacritty (as per WM_CLASS) and on x11 it works perfectly as soon as I set the gsetting, but reloggint into wayland it has no blur, so that's why I assumed that xprop was used internally to determine what apps to apply blur to.

CorvetteCole:
hey @aunetx is there anything here in particular you'd like me to have a go at? I'm going to have quite a bit of free time for the next week or so, looking for a project!

CorvetteCole:
I can always in general just work on polishing this PR as well

aunetx:
> hey @aunetx is there anything here in particular you'd like me to have a go at? I'm going to have quite a bit of free time for the next week or so, looking for a project!

Hello @CorvetteCole, sorry for the delay, as you may see I am quite busy but I try to find some time to continue this PR!

For the moment, the main thing to do is implementing the preferences UI, if you want you can spend some time doing this -- although this is pretty hard for the moment, as I have not finalized how I would like the UX to be (I still wonder about how I should place some buttons).

Except this (which I will work on anyway, and I begin to understand libadwaita so I should do something ok), you may want to try working on :
- #170, a big point for the extension; it is mostly working but as you can see testing it, there are some problems on `enable`
- #171, this would be super! It should not be too hard, but you may want to restart from scratch using `preferences-rewrite` (this PR) as a base, because it is probably less buggy and it will permit me to merge #171 without adapting it to #180
- #165, the same as before (you may want to start from scratch using the code I already implemented), it may be quite hard but should be done at some point :)

CorvetteCole:
thanks for the to-do list. I'll take a look this evening. Might take a look at #171 because I have multiple monitors and can easily test it. If you're okay with me doing some UX design I've got a bit of a knack for that so I might give that a go as well! We'll get this done

aunetx:
@CorvetteCole if you have the time, could you test this PR and see if everything works? It is now fully ready :)

bertin0:
I still have issues with blur not sticking on wayland as i mentioned above.

Demonstration (Tried using the new preferences on gnome 42 too but have reverted to 41 since):

https://user-images.githubusercontent.com/30109573/161449444-328e4d03-6c03-470c-97ba-5e5c740507dc.mp4

Basically it seems that the blur only gets applied to the windows open when the whitelist is first set, and that it doesn't properly check new windows.

PR #184, opened 2022-01-29
Fix some issues with dynamic blur and dash-to-panel

a2leexx:
Mutter is compositing window manager so it stores a texture buffer for each widget. For this reason widgets above a blurred background will not emit paint signal in some cases (because the compositor doesn't need to redraw them). This merge request fix issue by ensuring that the widget with the blur effect applied will be repainted entirely when some part of it is redrawn.

This merge request also contains small fix for problem with dynamic blur and dash-to-panel extension that occurs after locking screen.

### DIFF
--- a/src/dash_to_dock.js
+++ b/src/dash_to_dock.js
@@ -180,7 +180,7 @@ var DashBlur = class DashBlur {
             } else if (this.prefs.HACKS_LEVEL.get() == 2) {
                 this._log("dash hack level 2");
 
-                this.paint_signals.connect(dash, this.effect);
+                this.paint_signals.connect(background, effect);
             } else {
                 this.paint_signals.disconnect_all();
             }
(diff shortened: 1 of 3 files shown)


aunetx:
That looks really cool, thanks a lot! I will test this later today :)

a2leexx:
The latest commit fixes the issue that appears after using the overview with the vertical dock (in panel mode). Idea behind the latest two commits is that widget repainting caused by queue_repaint() also emit a signal, so it's needed to skip one or two signals.

aunetx:
That is amazing, thanks a lot!! I really did not know that repaints only occurred per-widget and that the compositor blended them together (well, I'm dumb, it's in the name... compositor), so I though that if two widgets were superimposed and one of the two was changed, the other was repainted as well!

Thanks to your addition, more "dynamic" blur may be really better, for example for the upcoming applications blur :)

Thanks a lot really!

PR #186, opened 2022-02-04
Bump to 42 since it works there

shemgp:
Tested using Ubuntu Jammy + GNOME PPA with gnome-shell 42.

Prefs works too.

### DIFF
--- a/src/metadata.json
+++ b/src/metadata.json
@@ -3,7 +3,8 @@
     "name": "Blur my Shell",
     "shell-version": [
         "40",
-        "41"
+        "41",
+        "42"
     ],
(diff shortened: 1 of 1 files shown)


timocapa:
No, it's been broken on main because panel corners have been removed. It'll fail to load with "Main.panel._leftCorner is undefined", probably the same with the right corner.

You could probably easy patch it by removing those functions, but it might break backwards compability.

aunetx:
Yes, due to the recent changes, this won't work with gnome 42. I will add "some" support for gnome 42 soon (by just verifying that corners exist before blurring them), but final support will only come with #180, which should use libadwaita for the preferences.

aunetx:
I merge this as #189 got merged too, will push to eog

PR #189, opened 2022-02-16
Fix missing panel corners in Gnome Shell 42

ariktaurendil:
Remove the workaround of hiding Gnome Shell panel corners that prevent this extension to work in Gnome Shell 42, since those panel corners were removed in Gnome Shell 42.beta.


aunetx:
Thank you very much, was planning to do this but this saves me some time! Will release for gnome 42.beta soon

PR #196, opened 2022-03-09
Fix _prepareWorkspaceSwitch is undefined

Schneegans:
The `disable()` method of the `OverviewBlur` can be called without a prior call to `enable()`. This can corrupt the prototype of the `WorkspaceAnimationController` if the extension is disabled while the overview blur is not enabled.


aunetx:
Sorry, I did not see this before, merging it thanks!

Issue #224, opened 2022-04-17
Error loading extension after locking screen


pesader:
Whenever my laptop goes to the lockscreen (either by Suspend or Lock) the "Overview components style" preference is ignored and the components go back to their original opaque appearance.

In addition to that, Extension Manager shows that the Blur My Shell could not be loaded (although everything seems to be working, except the Overview components styling). The error it displays is "this.effect.actor is null".

Logging out and logging in again brings everything back to normal.

| Opaque overview components | 
| ----------- |
| ![image](https://user-images.githubusercontent.com/65264536/163696642-d4448fd8-f736-41e0-97d3-6ac2861e7715.png) | 

| Error loading extension | 
| ----------- |
| ![image](https://user-images.githubusercontent.com/65264536/163696655-d283694f-f887-46a0-9a63-3dd125bef7fd.png) | 

And thanks a lot for developing Blur My Shell. I requested the overview components styling back in issue #174 and you made it look better than I imagined it. Great work! :star_struck:

retrixe:
I'm having the same issue, all the blurs are intact, but the dash and search cards are broken.
![image](https://user-images.githubusercontent.com/6884828/163725907-92ee6cf4-e8c5-4b21-857b-5851707c1a9b.png)
![image](https://user-images.githubusercontent.com/6884828/163725929-3cffbca6-d11c-4c73-9b18-43e50a5926e0.png)

zayats1:
> I'm having the same issue, all the blurs are intact, but the dash and search cards are broken (dash screenshot is in original issue).
> 
> Issues: ![image](https://user-images.githubusercontent.com/6884828/163726023-efc1232c-1efa-4c01-9a8d-de59d226e14b.png) ![image](https://user-images.githubusercontent.com/6884828/163725907-92ee6cf4-e8c5-4b21-857b-5851707c1a9b.png)
> 
> The error: ![image](https://user-images.githubusercontent.com/6884828/163725929-3cffbca6-d11c-4c73-9b18-43e50a5926e0.png) ![image](https://user-images.githubusercontent.com/6884828/163726012-ceb8f585-3e47-42fc-8865-9e91d45db54f.png)
> 
> I was able to patch this by editing `~/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/panel.js` line 250 like so:
> 
> ```diff
> -        if (this.prefs.PANEL_STATIC_BLUR.get())
[...]

PR #226, opened 2022-04-22
Fix error loading extension after locking screen

pesader:
This commit fixes issue #224, opened by myself.
Credit goes entirely to @retrixe, who figured out the solution.


ysolis:
#228 (it's a dup)

aunetx:
Thank you very much!

PR #230, opened 2022-04-29
Blur, but in color 

JugadK:
So currently the extension kinda overrides shell themes with the blur, and if you try to make the overview background and other components transparent in the theme it leads to alot of bugs (especially the overview background), So I made a patch blends a color into some of the blurred components.

It uses the Clutter.ShaderEffect which allows us to apply any glsl shader to clutter actors, the glsl shader can only be compiled once per ShaderEffect , you need to create a new shadereffect if you want to change the effect so I had to implement it for certain objects differently than the blur effect.

Since gnome does not really support theming I understand though if wan't to keep your extension closer to a "vanilla" gnome experience (with blur!)

This was my first time with gnome extensions so if I made any really dumb mistakes feedback is appreciated. 

image

image

image

image

### DIFF
--- a/resources/ui/coloreffect.ui
+++ b/resources/ui/coloreffect.ui
@@ -1,0 +1,133 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<interface domain="blur-my-shell@aunetx">
+  <template class="Color_Effect" parent="AdwPreferencesPage">
+    <property name="name">Color_Effect</property>
+    <property name="title" translatable="yes">Color Blur</property>
+    <property name="icon-name">color-symbolic</property>
+
+    <child>
+      <object class="AdwPreferencesGroup">
+        <property name="title" translatable="yes">Color blur</property>
+        <property name="description" translatable="yes">Change the color of the blur</property>
+        <property name="header-suffix">
+          <object class="GtkSwitch" id="color_blur">
+            <property name="valign">center</property>
+          </object>
+        </property>
+
+       <child>
+          <object class="AdwActionRow">
+            <property name="title" translatable="yes">Red</property>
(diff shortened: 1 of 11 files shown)


aunetx:
Thank you very very much for this addition!

I really like the idea and the implementation, but I should tweak some things before merging this -- I will work on this soon, as I don't have any time right now!

Thank you very much, that looks very promising :)

JugadK:
aight thanks! if its anything specific you want me to add I got time to burn at the moment so i'd be willing to attempt it.

aunetx:
I'm trying to review this to possibly merge it, but I might make quite a lot of changes, including:
- using `Clutter.BrightnessContrastEffect` instead of a custom shader; it is actually less precise (this effect does more than just mixing two colors), but is much more powerful (allows to increase and decrease the brightness for example, something I really wanted to do for quite a long time) and there is faster (as there is no need to compile it)
- change the preferences to use a real color picker instead of separade red/green/blue values
- maybe tie each effect to each component, for example to enable it for overview and panel but not the rest? I will see about it

Anyway, sorry for this, I might change some things from this PR; it you prefer to keep it this way for you, tell me and I will do the changes in another pull request so I don't touch to your own repo!

JugadK:
Yeah those sound great, athough I learned there is a way to get around the recompiling by using set_uniform_value() for the shader, but yeah those changes sound good.

aunetx:
I merged this in #246 instead, sorry for this! I really did not manage to rebase your work on the one I did just before, so instead I rewrote it -- that permitted me to add the preferences to each widget, and I finished it by adding a noise effect too.

Tell me if that is good for you, and thank you very much again!

Issue #141, opened 2021-11-05
Allow to change panel background colour


jmaris:
Hi,

When I enable panel blur in the extension, my panel (which is usually a semi-transparent white) becomes completely transparent with a blur effect applied. I have set brightness to 100%

Is this normal behavior, and is there any way to keep the colour of my panel ?

Tygrain:
Did you try to uncheck this checkbox?
![image](https://user-images.githubusercontent.com/50624811/140604376-38a9ab94-f483-4530-b96d-cb4b2f1f76dd.png)

jmaris:
Hi Tygrain, When I uncheck this box it removes blur entirely, my goal is not to remove blur, but rather to have blur but not remove the colour of the panel.

In short: my theme has a semi-transparent white panel, i want the panel to stay semi transparent and white, but for blur to be applied without getting rid of the semi-white colour.

aunetx:
I think I understand your request, but I can't do a lot of things right now, except giving you a quick hack! But for that, could you tell me the version of gnome you use? (so I can know which version of the extension you have)

jmaris:
Hey, thanks or the reply, i'm currently running on 40.5 (Manjaro)

aunetx:
You can try changing the file `.local/share/gnome-shell/exrensions/blur-my-shell@aunetx/stylesheet.css`, and remove the part:

```css
 ​#​panel​.​transparent-panel​ { 
 ​    ​background​:​ transparent; 
 ​    ​transition-duration​:​ ​500​ms​; 
 ​}
```

And then reload the shell, but this change will only last until next extension's update

jmaris:
Thank you ! Should I leave this issue open or close it ?

aunetx:
You can leave it open for others to see it, I will need to add a config option anyway :)

PR #234, opened 2022-05-04
Allow changing panel background color

RichardLuo0:
#141

### DIFF
--- a/resources/ui/panel.ui
+++ b/resources/ui/panel.ui
@@ -68,6 +68,29 @@
 
         <child>
           <object class="AdwActionRow">
+            <property name="title" translatable="yes">Panel background color</property>
+            <property name="subtitle" translatable="yes">Customize panel background color. Leave blank to use theme color. Ref to https://docs.gtk.org/gtk3/css-overview.html#colors</property>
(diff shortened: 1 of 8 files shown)


aunetx:
Hello, and thank you very much for your pull request!

I tried to merge this, unfortunately I had to change quite a lot of things, and used another merge request instead (I still don't understand how to rebase pull requests properly...): #245.

Unfortunately, I ran over a lot of bugs, when using your own pull request or the one I made; as the background-color won't be changed easily using `Main.panel.style` (there are a lot of conflicts and not easily resolvable, except if I patch the entire overview class -- but that's too much for such a simple feature), and using `Main.panel.set_background_color` (as I tried in my PR) instead is really great for having something that works every time, but lacks transition...

I will think about it, and either:
- I add a handmade animation (using loops and timeouts), which is horrific
- or I simply don't add this feature for the moment, and instead continue to work on #234 -- which will allow to more or less recolor the panel, if I do it correctly.

I think I will go with the second option, but tell me if there is another way :)

aunetx:
This was finally merged with #246, thank you very much!

PR #235, opened 2022-05-05
Update the blur when wm-class is changed

io12:
Some whitelisted programs (for example `xfce4-terminal` and `emacs`) have their windows blurred when the extension is first activated, but new windows are not blurred. I think this is because they set their wm-class after creating the window.

Steps to reproduce:
* Disable Blur my Shell extension
* Enable application blur and add `xfce4-terminal` to whitelist
* Open `xfce4-terminal`.
* Enable Blur my Shell
* Now `xfce4-terminal` should have a blur behind it
* Close and reopen `xfce4-terminal`
* The new window has no blur

This PR fixes this by watching for wm-class changes.

One problem that I haven't figured out is the blur rectangle is too large. Here's a screenshot showing this.

image

This problem also only happens for new windows. If a window is already open when Blur my Shell is enabled, the blur is the right size. Any idea what this could be?

### DIFF
--- a/src/components/applications.js
+++ b/src/components/applications.js
@@ -106,18 +106,20 @@ var ApplicationsBlur = class ApplicationsBlur {
             this.window_map.delete(pid);
         });
 
-        // update the blur when the mutter-hint is changed
-        this.connections.connect(
-            meta_window,
-            'notify::mutter-hints',
-            _ => {
-                let pid = meta_window.blur_provider_pid;
-                this._log(`mutter-hint changed for pid ${pid}`);
+        // update the blur when mutter-hint or wm-class is changed
+        for (const prop of ['mutter-hints', 'wm-class']) {
+            this.connections.connect(
+                meta_window,
+                `notify::${prop}`,
+                _ => {
+                    let pid = meta_window.blur_provider_pid;
+                    this._log(`${prop} changed for pid ${pid}`);
 
-                let window_actor = meta_window.get_compositor_private();
-                this.check_blur(pid, window_actor, meta_window);
-            }
-        );
+                    let window_actor = meta_window.get_compositor_private();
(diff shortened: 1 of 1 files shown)


JugadK:
I kinda looked into this and I think its because the blurring of new windows is suppose to happen on the signal connect on line 29 of applications.js but I think the syntax is wrong? if you do global.display.connect("window-created", () => {}) that runs the function each time a new window opens.

io12:
Also, this PR seems to only fix anything on Wayland. In Xorg there is no difference.

aunetx:
Thank you very much for your work!

I merge this right now, but that's weird if xorg still does not work... I will do some tests when I can, maybe under xorg the mutter hint change does not produce the same signal..?

aunetx:
And for the window size, to be honest I don't really know... Maybe `update_all_windows` does something I should do directly in `track_new` (which is called directly when a new window is created while BmS is already running)... But it does not seem so :/

Maybe I simply do not let time to the window to correctly set its size before the blur is called on it (a race condition I guess)... I will try to simply put a setTimeout to see if something changes!

io12:
> that's weird if xorg still does not work... I will do some tests when I can, maybe under xorg the mutter hint change does not produce the same signal..?

Sorry, I was confused because the wm-class was different in Xorg and Wayland for the programs I was testing. It looks like Xorg didn't have any problem to begin with. Xorg didn't have the "new windows not blurred" problem before the PR, and it doesn't have the "blur rectangle too large" problem after. I guess the programs set wm-class immediately on Xorg but not Wayland.

> Maybe I simply do not let time to the window to correctly set its size before the blur is called on it (a race condition I guess)... I will try to simply put a setTimeout to see if something changes!

I think this makes sense. Destroying and recreating the blur on window resize seems to fix it too, although maybe that isn't a good permanent fix?

PR #237, opened 2022-05-08
Complete translations support

oscfdezdz:
Since translations files are required as `.mo`, I thought using `gnome-extensions` command would make things easier for translation contributions as those files are already compiled and cannot be edited.

`gnome-extensions` takes care of compiling and placing translations files in their directories (`locale/<lang>/LC_MESSAGES/blur-my-shell@aunetx.mo`) and schemas.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,27 +1,25 @@
 NAME = blur-my-shell
 UUID = $(NAME)@aunetx
 
-.PHONY: build pkg install remove clean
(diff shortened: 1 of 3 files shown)


aunetx:
Thank you very very much! That was something I wanted to do for quite a long time, but I did not have the knowledge to do it properly!

Merging this right now, as well as the Spanish translations -- I will add French ones very soon :)

PR #240, opened 2022-05-10
Fix bug where blur rectangle is too big

io12:
This fixes the blur rectangle size bug described in https://github.com/aunetx/blur-my-shell/pull/235 by updating the blur actor's size constraints whenever the window's size changes.


aunetx:
Thank you very much, that's much cleaner than the previous code too :)

Issue #218, opened 2022-04-12
add noise texture option 


fadihajjej:
blur might look bad on older screens with lower contrast ratios due to the very small incremental hue changes, Windows combats this with some noise on to of the blur so it evens it out and makes it look smoother also noise on top of blur might look nice even on high end screens.

aunetx:
That's a good idea! That will require quite a lot of work, but I would love to provide this as an option indeed :)

JugadK:
Do you have an example of how much noise they use? It should not be too hard to implement with something like #230, though I feel like you would want to redo how the effects are added to make it easier to add and swap more effects in the future.

aunetx:
Done :)

fadihajjej:
amazing

PR #246, opened 2022-05-15
Add color and noise to the blur effect

### LINKED ISSUE #141: Allow to change panel background colour

aunetx:
This is a rewrite of #230, and features:
- an option to select the color of the blur, for each component (similar to sigma and brightness)
- a dynamic color effect, which does not need to be reloaded each time a preference is changed
- the preference is a `Gtk.ColorButton`, whose alpha slider controls the blend

It will also replace #234 (because the panel color will be changeable), and thus will replace #245 too. It will close #141 by the way.

The "only" bits missing are:
- [ ] dash-to-dock color implementation
- [ ] applications color implementation
- [x] the panel is not being correctly updated when the new color has a blend of zero
- [ ] add a global switch, as it may deteriorate performances for some user, so the ability to disable it for every components at once may be welcomed
- [ ] lockscreen looks weird, I think I did not add the effect to the right widget so the whole screenshield is colored
- [x] whenever an effect is added to the appfolders, it becomes very dark
- [ ] a lot of tests are needed, as this is quite a big feature

Thank you, @JugadK and @RichardLuo0; and sorry for not having merged your requests instead! This work is really yours, I simply tried to make something more homogeneous :)


aunetx:
This will also now close #218, as I added a noise effect (or the beginning of it). It is currently applied to the overview and panel, but I will add it to the rest too; and add the preferences for it.

I may want to change the way the effects are added, to permit the user to "drag" them and reorder the way they are applied, but that looks too much work for little to no added functionality...

aunetx:
Noise effect now seems complete.

PR #265, opened 2022-05-27
A corner effect, working... when no blur is used.

aunetx:
The effect kinda works, but can't be used in conjunction to `Shell.BlurEffect`. This was the goal, obviously.

Next steps:
1. create a custom glsl effect for a gaussian blur effect -- or kawase, of fast blur, or anything that is good looking and fast enough (if anybody wants to provide something, I'm very very open...)
  - OR just get [the original from the horn of plenty GNOME shell code](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/src/shell-blur-effect.c), although I don't understand it... how is it not GLSL?
  - OR use and modify [the original blyr effect](https://github.com/yozoon/gnome-shell-extension-blyr/blob/master/blyr%40yozoon.dev.gmail.com/shader.glsl), if @yozoon is okay with that (at least I understand the code :p)
3. integrate it with the corners effect (which is kinda simple, just checking if the pixel is or not in the corner with the type of conditions I put here, and here we go)
4. integrate the noise and color effect with it (which would really be easy I guess)
5. change the way the extension asks for blur, to use this effect instead of `Shell.BlurEffect`

If done well, this could be approximately as fast as the default `Shell.BlurEffect`, or faster if the other effects are used with it, as it would not need to mix different effects.

However, this very probably won't work for any kind or dynamic blur, but at this point I'm not even asking for this... especially if #170 can be finally finished, as nearly all the important parts of the shell would then be appropriately blurred.

### DIFF
--- a/src/effects/color_effect.js
+++ b/src/effects/color_effect.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { GLib, GObject, Gio, Clutter, Shell } = imports.gi;
+const { GLib, GObject, Clutter, Shell } = imports.gi;
 const ExtensionUtils = imports.misc.extensionUtils;
 
 const Me = ExtensionUtils.getCurrentExtension();
--- a/src/effects/corner_effect.glsl
+++ b/src/effects/corner_effect.glsl
@@ -1,0 +1,42 @@
+uniform sampler2D tex;
+
+uniform float width;
+uniform float height;
+uniform float radius = 10;
+
+vec2 pixel_coord = cogl_tex_coord_in[0].xy;
+vec2 uv = pixel_coord * vec2(width, height);
+
+bool is_corner_to(float x, float y) {
+    return distance(uv, vec2(x, y)) > radius;
+}
+
+void main() {
+    vec4 pixel_color = texture2D(tex, pixel_coord);
+    float opacity = 1.;
+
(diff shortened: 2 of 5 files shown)


JugadK:
So this was actually what I originally wanted to do instead of a color effect, but I stuck to that since I hadn't messed with gnome extensions before, anyway before finals I was trying to do something like this  (was struggling with the corners thanks for that) and ill just put some info that might be helpful to others

This is a glsl implentation that is in mutter uses, it does alot of optimizations with clutter and the syntax is different than gnome-shell glsl shaders but its the one pass blur that you see in alot of online docs, 
https://github.com/GNOME/mutter/blob/main/clutter/clutter/clutter-blur.c
I tried implementing it but never got it work

This one has a box-blur already implemented for gnome shell using glsl, though it does not look as nice as Gaussian blur
https://github.com/GNOME/mutter/blob/main/src/tests/clutter/interactive/test-cogl-shader-glsl.c

This relates some of the cogl glsl shader keywords to the normal glsl shader keywords you will see 
https://developer-old.gnome.org/cogl/stable/cogl-Shaders-and-Programmable-Pipeline.html

Though that blyr one looks nice, I might try to see if I can add it into the corner effect tonight.

aunetx:
Superseeded by #533

PR #271, opened 2022-05-29
Try to fix #94

### LINKED ISSUE #40: Blur effect not working - multi monitor.
Hello, I am opening bug as per request.

I have attached 2 screenshots showing an issue relating to setting a sigma value. When it is above 0 then you do not see a background on secondary monitor, at least for me.

Please see 2 screenshots:

Kind regards
Peter
Screenshot from 2021-04-21 15-16-33
Screenshot from 2021-04-21 15-16-26

### LINKED ISSUE #94: On Wayland with multiple displays, the overview background on the primary display is gray sometimes
Quick screen recording showing this: https://nextcloud.laurinneff.ch/s/rdr8nDMZWeQBqE9 (uploaded on another site because GH doesn't support .webm)

On Xorg, it always works as it should. Non-primary displays also work every time. It only happens on the primary display, when multiple displays are connected, and if it's running on Wayland

(I haven't tried it with three displays, maybe the behavior is different then)

aunetx:
Needs testing from multi-monitors people :)

### DIFF
--- a/src/components/overview.js
+++ b/src/components/overview.js
@@ -40,8 +40,8 @@ var OverviewBlur = class OverviewBlur {
             _ => {
                 if (GLib.getenv('XDG_SESSION_TYPE') == "wayland" &&
                     Main.layoutManager.monitors.length > 1) {
-                    this.repaint_blur_effects();
-                    //this.toogle_actors_visibility();
+                    //this.repaint_blur_effects();
+                    this.toogle_actors_visibility();
                 }
             }
         );
@@ -222,7 +222,7 @@ var OverviewBlur = class OverviewBlur {
     toogle_actors_visibility() {
         Main.layoutManager.overviewGroup.get_children().forEach(child => {
             child.hide();
-            child.show();
+            setTimeout(_ => child.show(), 10);
         });
     }
 
--- a/src/components/overview.js
+++ b/src/components/overview.js
@@ -38,9 +38,11 @@ var OverviewBlur = class OverviewBlur {
             Main.overview,
             'showing',
             _ => {
-                if (GLib.getenv('XDG_SESSION_TYPE') == "wayland" &&
-                    Main.layoutManager.monitors.length > 1) {
-                    this.toogle_actors_visibility();
+                if (
+                    GLib.getenv('XDG_SESSION_TYPE') == "wayland" &&
+                    Main.layoutManager.monitors.length > 1
+                ) {
+                    this.update_backgrounds();
                 }
             }
         );
@@ -210,15 +212,6 @@ var OverviewBlur = class OverviewBlur {
                 group.remove_style_class_name("bms-overview-components-dark");
                 break;
         }
-    }
-    
-    toogle_actors_visibility() {
-        Main.layoutManager.overviewGroup.get_children().forEach(child => {
-            if (child.constructor.name === 'Meta_BackgroundActor') {
-                child.hide();
-                setTimeout(_ => child.show(), 10);
-            }
-        });
     }
 
     set_sigma(s) {
--- a/src/components/overview.js
+++ b/src/components/overview.js
@@ -40,7 +40,6 @@ var OverviewBlur = class OverviewBlur {
             _ => {
                 if (GLib.getenv('XDG_SESSION_TYPE') == "wayland" &&
                     Main.layoutManager.monitors.length > 1) {
-                    //this.repaint_blur_effects();
                     this.toogle_actors_visibility();
                 }
             }
@@ -212,17 +211,13 @@ var OverviewBlur = class OverviewBlur {
                 break;
         }
     }
-
-    repaint_blur_effects() {
-        this.effects.forEach(effect => {
-            effect.blur_effect.queue_repaint();
-        });
-    }
-
+    
     toogle_actors_visibility() {
         Main.layoutManager.overviewGroup.get_children().forEach(child => {
-            child.hide();
-            setTimeout(_ => child.show(), 10);
+            if (child.constructor.name === 'Meta_BackgroundActor') {
+                child.hide();
+                setTimeout(_ => child.show(), 10);
+            }
         });
     }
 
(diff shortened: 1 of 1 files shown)


aunetx:
Would close #40 too, and maybe some other issue...

bdaase:
I gave this a shot. Unfortunately, however, the problem with multi monitors having a gray background persists…

RaameshB:
Alright tested this, used gh pr checkout 271 after cloning the repo and then did a make install after uninstalling the extension

unfortunately the issue was not resolved

aweirddudee:
Just tested doing the same thing as Neon-44 - although it hasn't made it persistent, it changed my primary display from being constantly default to on/off with the blur effect from the extension - more often blurred than grey. Hopefully the lead from switching monitors off and on in sequence can go somewhere. :)

crstmkt:
Did the same thing as Neon-44. This has fixed the issue for me so far (using v42 of the extension). Thanks for sharing!

Mrcuve0:
Whish I could say the same. Tested with v43.
Setup: I have my laptop on the right of my 24'' monitor (which is set to be the primary display, see picture)

image

I did the same as suggested by Neon-44:
1. Turned off (from settings) my laptop display and turned on again;
2. Turned off (from settings) my 24'' external monitor and turned on again;
3. Mirrored both screens
4. Joined them
5. Moved my external monitor on the left of my laptop display
6. Set my external monitor as primary display

Step 5. and 6. are necessary as Settings sets by default my external monitor on the right of my laptop (I want it to be on the left side)

Henry-Hiles:
> i don't know if it is because you fixed it, but it resolved itself for me after i:
> 
>     1. in the settings: turned the screens off and back on one after the other
> 
>     2. made the screens mirror, then joined again.
> 
> 
> it has been fixed and persistent through reboots and even a system update (fedora 36) idk if this helps you, but i hope it does :) also thanks for the work you're doing here, i really appreciate it, this is my favourite extension by far, and i personally think it should be built into GNOME by default :)

This, unfortunately, did not work for me.

Henry-Hiles:
> Whish I could say the same. Tested with extension @ v43, Fedora 36, GNOME 42.4 Wayland Setup: I have my laptop on the right of my 24'' monitor (which is set to be the primary display, see picture)
> 
> image
> 
> I did the same as suggested by Neon-44:
> 
>     1. Turned off (from settings) my laptop display and turned on again;
> 
>     2. Turned off (from settings) my 24'' external monitor and turned on again;
> 
>     3. Mirrored both screens
> 
>     4. Joined them
> 
>     5. Moved my external monitor on the left of my laptop display
> 
>     6. Set my external monitor as primary display
> 
> 
> Step 5. and 6. are necessary as Settings sets by default my external monitor on the right of my laptop (I want it to be on the left side)
> 
> The issue is still there

Same here, perhaps its because of us moving he external monitor?

srdante:
A strange fact about my setup. I have a notebook + 2 external monitors (34' and 27'). I usually use the notebook folded and only with the two external monitors enabled.

Using this way, the bug happens on the primary monitor (34').

If I open my notebook to use the three monitors (notebook one + 2 externals), the bug disappears and the extension works well on all monitors!

AshtakaOOf:
The "fix" from Neon doesn't work at all.
So anything new on this?

aunetx:
> The "fix" from Neon doesn't work at all. So anything new on this?

Nothing new, sorry. I don't even have a secondary monitor with me for the moment, so except if someone provides a true fix via another pull request, I can't do anything.

Neon-44:
oh yeah, forgot to post:

after some more testing, i found out, that all the steps i listed weren't really necessary. for me it also worked if i just moved the screens in the settings app and applied.

not quite sure if it has something to do with how the screens are positioned exactly, but i am tending to "it doesn't have anything to do with it" based on testing

@AshtakaOOf

AshtakaOOf:
Doing that made things even worse, now it doesn't even blur the background anymore...

Fedora 37
Gnome 43.2 
Mutter 43.2

@Neon-44

Neon-44:
hmm, that's weird

doing this 1-3 times fixed it permanently for me

i am sorry, can't help you currently, as i am having [this](https://github.com/NixOS/nixpkgs/issues/203355) Problem and can't even boot Wayland in GNOME

AshtakaOOf:
It now works 3 times out of 5 after doing that but not 100% of the time. :disappointed:

PlayLikeThat:
Seems like it has something to do something with the placement of the monitors.
If I manage to place it exactly in the middle, then it works 100% out of 40 tries (Hot Corner + Super)

When it is not centered, then it behaves like @AshtakaOOf describes, and snapped to the top or bottom disables it completely.

https://user-images.githubusercontent.com/73217346/212977579-6a70b835-3116-4be9-824b-07a5ead7caec.mp4

AshtakaOOf:
The problem is the fact that they are centered, but they are both 1080p so it doesn't work. :(

srdante:
I created an issue at the gnome-shell repository: https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/6370

aunetx:
Hi @srdante, thank you very much for your implication! I'm sorry I did not do this before, but I hope this could get fixed by the gnome team

michelchalmers:
I think I have found the problem causing the blur malfunction on multiple monitors. I am using Blur My Shell on Wayland with 2 Dell monitors. I have noticed that using Blur My Shell with Dash to Dock is not a problem unless you use the application icon from Dash-to-dock. when you click on the application icon, the blur disappears, but if you use the meta key (windows) instead, it works fine. Windows key+A will work fine as well showing the blur. As I said, it is only when I use the application icon from Dash-to-dock that I experience the intermittent blur problem. So, I have disabled the application icon from Dash-to-dock and I am only using meta, or meta+A....So, if you disable the application icon in Dash-to-dock, or do not use it, you will not have the problem


https://github.com/aunetx/blur-my-shell/assets/145332847/2c9ca486-36e3-45a5-8303-378ded076030

Issue #90, opened 2021-06-15
Feature request: make top panel blur dynamic.


Antoshka1488:
Blur when there is no maximized window and window on top. Solid in other cases.

Ferryistaken:
Yes please this would make the extension way more enjoyable to use

aunetx:
I will see what is possible to do, my other extension (Smart Transparent Topbar, still not compatible with gnome 40) did this so implement it would not be too hard.

aunetx:
When you talk about a solid topbar, do you think about a "normal" topbar (black, with adwaita), or eg a transparent topbar? The first option is a lot easier, because I would not have to manage the transitions, but the second option seems more enjoyable to me (but ofc, if I implement the transparent option, it would be optionnal)

What would you want? Just solid, or "solid-or-transparent"?

Ferryistaken:
To me just a normal topbar when an application is maximized would be fine, but if you think that it would be more enjoyable to make it normal-or-transparent I think that that is fine too, since I would still be able to make it solid.

aunetx:
Ok thanks!
Well, I will first try to implement the behaviour with the normal topbar colour, and after maybe see for something else (although... this may actually not be my job, as another theme/extension could change the topbar colour too) :p

Will try to begin tomorrow if possible!

Ferryistaken:
Perfect, thanks for your effort in making this extension

aunetx:
I will work on this as soon as I merge the branches `window-list` and `blur-onscreen-keyboard` (first should be very quick, as everything works (see #103), but the second needs some more work, although I could add this later as no issue is opened about that).

aunetx:
`window-list` is done now, and `blur-onscreen-keyboard` is too hard to be finished by me alone... So I begin to work on this, but idk when I will have something that works :p

aunetx:
I began the work in branch `dynamic-blur`, but I have a hard time understanding why I can't get my extension to update itself when its position is changed...

I will need some time to get is done, but at least I only have bugs to fix now!

aunetx:
This will be fixed very soon by #272 :)

aunetx:
Fixed :)

PR #272, opened 2022-05-29
feat: disable panel blur when window is close to panel

### LINKED ISSUE #239: Disable top-bar blur in maximized mode
I'm not sure if this has been asked previously or not, sorry for that, but..

**An option to disable top-bar blur when a window is maximized is quite promising.**

This is possible as Ubuntu Dock as well as Dash to Panel extensions already have this feature.

Personally, on F36, I've the panel blur disabled because it looks wrong when a window like browser is maximized.

TomBursch:
Closes #239

This feature tracks all windows and whenever one of them is near the panel, the transparency is removed.

Thanks for your great work, love the project!
If there is anything that needs change, feel free to let me know.

### DIFF
--- a/resources/ui/panel.ui
+++ b/resources/ui/panel.ui
@@ -50,6 +50,21 @@
             </child>
           </object>
         </child>
+
+        <child>
+          <object class="AdwActionRow">
+            <property name="title" translatable="yes">Disable in fullscreen</property>
+            <property name="subtitle" translatable="yes">Disables the blur from the panel when a window is near it.</property>
+            <property name="activatable-widget">unblur_in_fullscreen</property>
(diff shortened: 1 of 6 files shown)


aunetx:
Wow, thank you very much! Looking into this :)

aunetx:
I will change some things about it, but the base is here!

Notably, I would like to allow to fully select the behaviour in each situation: have options for [transparent / blurred / normal] topbar, when [window touches it / no window touches it / overview]

But I think I will do this later, and for the moment simply change some labels in the preferences and some variable names, and merge this!

By the way, this superseeds #165 and closes #90 too. Thank you very much!

aunetx:
Thank you very much!

PR #293, opened 2022-07-09
minor grammar corrections and use === instead of == where applicable to prevent type coercion

CorvetteCole:
Some grammar things were annoying me while I worked on improving applications blur, so I went through the whole project


CorvetteCole:
(I've got another merge request on the way to add a proper application blur selection menu)

aunetx:
Thank you!

Issue #198, opened 2022-03-11
Command to start application with blur 


andrewaclear:
How can we add `blur-provider=${sigma-value}` to `_MUTTER_HINTS` as we start an application i.e. is there a terminal command when starting an application? For instance `<<something>> gnome-terminal` to start the terminal with blur behind it?

carlosecp:
Same question here. I don't know much about how properties work in X windows but would love to know if there is a way to launch an application with the blur property set.

aunetx:
Hello, to be honest I don't know at all! About this, @CorvetteCole is much better than me :)

If you really want to do this, but is a better way (without needing to change the way you start you app of anything), you can use #180 (branch `preferences-rewrite`), this will allow you (from gsettings, I explained in the PR how to use it) to select the name of the app you always want to blur.

Ofc, this will be accessible through the extension preferences, hopefully with a nice-looking ui, but for the moment this is all I have :)

CorvetteCole:
ooh that's a good question. Let me look at the docs and I'll get back to you guys

andrewaclear:
thanks guys, but ya @aunetx that's true, what you are saying is even better!

CorvetteCole:
after exploring this I haven't found a good solution. I think I will build a small wrapper application that launches the target application with the blurring property set. For example `blur-app YOUR_APP_COMMAND`

aunetx:
@CorvetteCole I don't think it would be the best thing to do, as the user can select (with the whitelist, under gnome 42) which app to blur, and using mutter hint should only be used directly by the app itself.

So, once the applications blur selector is really well implemented (and not just the current text line I added in the prefs), the user would only have to touch to the extension preferences to blur the app (s)he wants: hence, the next step is to have a nice window selector, directly in the preferences.

CorvetteCole:
you're right, it would probably be better to focus on implementing that. The new preferences look great by the way. I'm going to try to help iron out the applications blur implementation and then maybe we can implement my really hacky corner blur workaround (but it works!). I've got a proof of concept I'll put on git soon for feedback

andrewaclear:
Thanks a lot guys. Ya I think selecting the apps that we want to be blurred in the settings is much better in the end, don't worry about the command

andrewaclear:
with the whitelist now this command should not be necessary

aunetx:
With #304 this will be selectable from the preferences directly

andrewaclear:
[...]

Issue #197, opened 2022-03-09
Option to blur all applications


ardnod:
Except: (manual setting, e.g. app1,app2...).

aunetx:
This will be fixed with #304.

PR #304, opened 2022-08-12
Add whitelist and blacklist preferences for application blur

aunetx:


aunetx:
Now simply needs the `enable all` behaviour extension-side

aunetx:
Will close #198 and #197

CorvetteCole:
this is awesome! I'm still working on an applications menu type search menu, we can use that to validate window names etc.

CorvetteCole (commented):
Looks great! I'll give it a test in a few

CorvetteCole (approved):
seems to work as expected looks good to me!

aunetx:
Thanks :)

CorvetteCole:
@aunetx I just realized that it seems a blacklist was enabled by default with blurring all applications a default option. Can you double check to make sure that isn't true? wouldn't want to reduce performance for all users by default

aunetx:
@CorvetteCole uh oh, nice catch! That's ofc not what I intended; just fixed this (I am working on multiple panels support in the same time so I will commit this asap)

Issue #295, opened 2022-07-14
Dash to Panel is not blurred when using the original panel too


SnoutBug:
When using Dash to Panel v50 and Blur my Shell v39 on Gnome 42 the blur is only applied to the top bar not the panel itself 
![image](https://user-images.githubusercontent.com/71790678/179006238-768c52d0-00c6-41ec-abfb-6ca2429510d9.png)

DTP Settings:
![image](https://user-images.githubusercontent.com/71790678/179007432-4714a36c-8445-44e2-9f92-0e102b403ffe.png)

BMS Settings:
![image](https://user-images.githubusercontent.com/71790678/179007982-c5f2854f-555c-483e-a73a-4a378a061ad0.png)
Does not work with either static blur turned on or off

If necessary I will attach some logs if you tell me where I can find them :D

aunetx:
Hello, this is due to the option "keep the original panel" in "personalisation" in Dasht to Panel preferences. BmS does not support multiple panels for the moment, so it will blur only the original one!

You can either disable this option, or wait until #171 is merged (I still have not worked on this, sorry) :)

SnoutBug:
Ah I see, thanks for the heads up

aunetx:
@SnoutBug if you want, you can try #306, tailor-made for you :)

SnoutBug:
Awesome, works like a charm on my ThinkPad! I will try this on my main machine later and report any issues should there be any.

aunetx:
Thank you very much! You can also try it with multiple monitors, but for the moment this should not work well

PR #306, opened 2022-08-17
Blur multiple panels when using dash to panel

### LINKED ISSUE #152: Blur multiple panels (Dash to panel)

aunetx:
Superseeds #171, will fix #152 and #295.

This basically works well, when using only one monitor.

I've not tested with multiple ones, it should work OK when using dash to panel, but it not dynamic though (when adding a monitor, its panel will not be blurred).

I should also add support for multiple panels without dash to panel if needed, it would be fairly easy but may depend on the extension used.

### DIFF
--- a/src/components/appfolders.js
+++ b/src/components/appfolders.js
@@ -255,6 +255,6 @@ var AppFoldersBlur = class AppFoldersBlur {
 
     _log(str) {
         if (this.prefs.DEBUG)
-            log(`[Blur my Shell] ${str}`);
+            log(`[Blur my Shell > appfolders]   ${str}`);
     }
 };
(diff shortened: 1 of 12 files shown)


aunetx:
I found a second monitor, that's insanely difficult to debug...

aunetx:
Now kinda works with multiple monitors, but it's way to hard for me to continue working on this :)

I this I will ship it like that ; it more or less breaks when using multiple dtp panel PLUS the original one, but at this point I don't know what to do.

It should at least work for simple and medium configurations, which is better than before.

aunetx:
(sorry @CorvetteCole to ask your help here, it's just to see if you can see some obvious error on my spaghetti code before I ship it to everybody)

CorvetteCole:
no problem, I'm more than happy to review. it is a good way for me to contribute when I don't have as much time to do my own projects. I will look at this very soon!

aunetx:
This finally looks quite OK, I will merge this right now to continue working on something else

Issue #312, opened 2022-08-21
Version 41 Error this.service is undefined 


TanawatJukmongkol:
The latest version of blur-my-shell randomly crashes, and when turn off blur-my-shell consistently, and I don't know how to debug it.

aunetx:
Hello, thank you for opening the issue. Could you give me the logs that happen after it crashes with `sudo journalctl /usr/bin/gnome-shell|tail -n 300`?

I think I may know where this could come from but I'm not sure.

tyn7:
I am also experiencing this issue. This extension crashes after locking the screen or waking from sleep. Here is the log from a cold boot, then locking/unlocking the screen
[log.txt](https://github.com/aunetx/blur-my-shell/files/9389168/log.txt)
:

Leo3418:
After enabling applications blur, I no longer encounter this issue.  If you don't want to use it, you may still enable it but turn off "Enable all by default".

![image](https://user-images.githubusercontent.com/14175175/185798549-cca18886-e2bf-4bc6-8a38-fef042864cdd.png)

P.S. I think this option is enabled by default, but if I were developing this extension, I'd be skeptical about turning on a beta feature for users by default, especially when I myself thought the beta feature was buggy from the developer's standpoint.

TanawatJukmongkol:
> Hello, thank you for opening the issue. Could you give me the logs that happen after it crashes with `sudo journalctl /usr/bin/gnome-shell|tail -n 300`?
> 
> I think I may know where this could come from but I'm not sure.

Tysm! Here's the log:
```
Aug 21 20:38:18 pop-os gnome-shell[3138]: st_widget_get_theme_node called on the widget [0x55d60147b030 StBin:insensitive ("")] which is not in the stage.
Aug 21 20:38:18 pop-os gnome-shell[3138]: st_widget_get_theme_node called on the widget [0x55d6015df2b0 Gjs_ui_panel_AppMenuButton.panel-button:insensitive ("")] which is not in the stage.
Aug 21 20:38:18 pop-os gnome-shell[3138]: st_widget_get_theme_node called on the widget [0x55d60147c700 StBin:insensitive "appMenu" ("")] which is not in the stage.
Aug 21 20:38:18 pop-os gnome-shell[3138]: st_widget_get_theme_node called on the widget [0x55d60147dd90 StBoxLayout.panel-status-menu-box:insensitive ("")] which is not in the stage.
[...]

PR #314, opened 2022-08-22
Fix applications service error

justperfection:
The service will be undefined when user don't use application blur.

Since `?.` is supported in 40 and higher, we can use that to avoid calling undefined properties.

Related to #312

### DIFF
--- a/src/components/applications.js
+++ b/src/components/applications.js
@@ -411,7 +411,7 @@ var ApplicationsBlur = class ApplicationsBlur {
     disable() {
         this._log("removing blur from applications...");
 
-        this.service.unexport();
+        this.service?.unexport();
 
         this.blur_actor_map.forEach(((_blur_actor, pid) => {
             this.remove_blur(pid);


TanawatJukmongkol:
Yup. Looks like that's the fix. I think in the long run, using monads to check for undefined or null, instead of if statements could be beneficial (that's just my thoughts on this).

TanawatJukmongkol:
Now that I think about it... If `this.service` is undefined, then this could mean there's some undefined behavior else where (which, if that's the case would make this only a work around). And not calling the `service.unexport` might lead to memory leak?

justperfection:
@TanawatJukmongkol
No, that's happening because the `enable` method hasn't been called before (as I explained in the commit).
And `?.` is just a quick fix because I didn't want to change the architecture.
Basically, when it's not enabled, the first line of disable function should `return`.

This error is not easy to find on review time, that's why I didn't catch it on ego review.

TanawatJukmongkol:
> @TanawatJukmongkol No, that's happening because the `enable` method hasn't been called before (as I explained in the commit). And `?.` is just a quick fix because I didn't want to change the architecture. Basically, when it's not enabled, the first line of disable function should `return`.
> 
> This error is not easy to find on review time, that's why I didn't catch it on ego review.

Ahh, I see. Thanks for the clarification. 😊

aunetx:
Thank you very much, I'm kinda dumb and forgot that I choose to disable every component, even when they haven't been enabled before... That's stupid and I will change this behaviour, but at least this will fix the crash very well :)

PR #325, opened 2022-08-26
Add basic per-window transparency setting

aunetx:
This may be too hard to maintain, and does not add a lot of good to the extension.


aunetx:
Linked to #329

PR #326, opened 2022-08-26
Add an option to change the opacity of the applications

aunetx:
This option is for the moment only general, it may be expanded later to be applied per-application.
The default application blur preferences has also been changed in order to minimize the blur darkness.

### DIFF
--- a/resources/ui/applications.ui
+++ b/resources/ui/applications.ui
@@ -23,6 +23,27 @@
 
         <child>
           <object class="AdwActionRow">
+            <property name="title" translatable="yes">Opacity</property>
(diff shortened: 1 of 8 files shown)


aunetx:
This finally completes a good application blur :)

cc @CorvetteCole

CorvetteCole:
there is one more thing on my list: better application protocol for
requesting blur (preferably protocol agnostic). current mutter hints method
is kind of, but really not meant for Wayland. I've got a few ideas though
(we have our own little dbus interface probably). Also to figure out
rounded blur corners... ugh. perormant solutions for that are very hard

looks great :)

PR #374, opened 2022-11-14
Fix multi monitors

aunetx:


Its-Juice:
If this is implemented into the main branch it will fix multiple issues regarding multi-monitor setups.

PR #379, opened 2022-11-23
add hungarian translation

PsyTheGuy:
Almost finished hungarian translation. The two remaining parts are:

#: customize-row.ui:92 general.ui:94
msgid "Noise lightness"
msgstr ""

#: customize-row.ui:93 general.ui:95
msgid "The lightness of the noise added to the blur effect."
msgstr ""

I left these because I couldn't find out what the option actually does. I tried to set it different values but couldn't see anything visually change. If you tell me what it does, I will finish the translation.

Cheers!

### DIFF
--- a/po/hu.po
+++ b/po/hu.po
@@ -25,7 +25,7 @@ msgstr "Alkalmazások"
 
 #: applications.ui:10
 msgid "Applications blur (beta)"
-msgstr "Alkalmazások elhomályosítása (beta)"
+msgstr "Alkalmazások homályosítása (béta)"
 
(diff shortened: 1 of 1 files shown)


aunetx:
Thank you very much :)

aunetx:
These options are about the colour of the dots added in the noise effect, quite subtle but you can see the effect if you set a strong noise intensity... I should not have put this in the preferences anyway, sorry for the hassle :/

aunetx:
(I merge this, if you want to add the remaining translations feel free to open another PR or continue pushing here)

Issue #402, opened 2023-02-03
Memory leak in overview blur


hcoggins:
## Description
I've noticed that showing the gnome overview or switching workspaces causes the allocated shared memory to increase when the background blur for the overview is turned on. Over the course of a working day, this can cause the allocated shared memory to be very high (upwards of 20GB on my machine), leading to performance problems. This memory leak does not occur when the background blur for the overview is turned off. 

I have observed this on two different machines and on both Wayland and X-org.

The shared memory can only be de-allocated by logging out in Wayland, or resetting the shell in X-Org (`Alt-f2 + r + Enter`).

## Demonstration

https://user-images.githubusercontent.com/40951916/216572079-e298af61-5a72-41a6-b82b-4b9efbf2d4f2.mp4

The above video shows how the memory leak occurs when blur is turned on. The plot shows the allocated shared memory of the system (output of `free | awk 'NR==2 { print $5 }'`) over time.

## System Info
**OS**: Arch Linux
**Kernel**: 6.1.9-arch1-1
**Gnome Shell**: 43.2
**Compositor**: Wayland

hcoggins:
After a bit of playing around with the source code, I think it has to do with the background-actors in `src/components/overview.js`. I'll give an update if I find out more.

DimitrisMilonopoulos:
Ah yes had the same issue. I disabled the extension completely.

I feel until another maintainer steps in, or main maintainers can allocate time.
This extension should be marked as unsuitable for use.

aunetx:
Hello @hcoggins, thank you for your detailed issue, I will try to figure out what is wrong... I never noticed it before, even though I checked regularly for such leaks using system monitor, sorry :/

Would you mind sharing the code you used to show the memory graph? I don't know well gnuplot, and that's way better than just looking at the number to test it :)
I keep you updated!

aunetx:
Could you try the PR #405 please? I think it should be fixed, but I'm not entirely sure (it does not increase the memory anymore for me, but it could just be something else)

inthreedee:
I gave it a quick test. Before the test, I saw my shared memory going up a few MB every time I went into overview or swapped desktops. After a reboot, it stopped happening though so I suspect uptime might also be a factor?

In any case, after the patch, my shared memory doesn't appear to be increasing anymore on Wayland or Xorg. But I haven't done exhaustive testing and only left my system up for a day so far.

hcoggins:
[...]

PR #405, opened 2023-02-10
Try to fix memory leak in overview blur #402

aunetx:


DimitrisMilonopoulos:
Still experiencing memory leak with this branch. I initially though it was fixed but after a while my mem usage was 92% of my 32GB machine.

src/components/overview.js
```diff
@@ -149,6 +149,9 @@ var OverviewBlur = class OverviewBlur {
         Main.layoutManager.overviewGroup.get_children().forEach(actor => {
             if (actor.constructor.name === 'Meta_BackgroundActor') {
```
DimitrisMilonopoulos, line 150: Not sure I know the code base, but are there different names other than 'Meta_BackgroundActor' ? 

Maybe that causes the leak.

aunetx:
This is superseeded by #480, which finally fixes this bug

PR #480, opened 2023-10-10
Add an effects manager for custom clutter effects to prevent ram bleedings

aunetx:
Does not work at all for the moment


aunetx:
Looks like I can at least remove blur effect-related things here, as the bug only affects noise and color effects (so definitely a bug in clutter), at least for me (I will need additional infos to be certain).

aunetx:
Fixes #402, superseeds #405

PR #495, opened 2023-11-10
Afrikaans language file

Leon2332:
Afrikaans language file. Copied and edited from the Dutch language file (nl.po)


aunetx:
Thank you very much!
By the way, for next updates, you may use weblate (so that all the translations are done in the same way, and because that's surely easier for you :D)

Leon2332:
Thanks.
Sorry, I only just realised more languages can be added to the list when I've registered on webplate. I'll get it right next time.

PR #531, opened 2024-02-22
Fix #377

### LINKED ISSUE #377: [BUG] Blur stays applied on the top panel after the system wakes up from suspend state
image

Blur stays applied on the top panel after the system wakes up from suspend state. But this isn't a problem when system is in locked state.

Expected:
Blur not being applied on top bar, just like how it is in lockscreen.

What happens:
The blur shows up. Which looks very out of place.

Delta1925:
This pull request adds the appropriate session-modes to `metadata.json` to keep the extension working after the system wakes up from a suspended state.

Adding `unlock-dialog` to the session-modes [should not affect the behavior of the extension while the user is logged in](https://gjs.guide/extensions/overview/anatomy.html#session-modes) and should therefore be safe to add.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -15,5 +15,9 @@
         "github": "aunetx",
         "kofi": "aunetx"
     },
+    "session-modes": [
+        "user", 
+        "unlock-dialog"
+    ],
     "version": 55
 }


aunetx:
Thanks! Have you tested it? Last time I tried this permission it worked better but not totally, and it did not get approved on extensions.gnome.org... But if it fully works I will try to convince them to publish it anyway! :)

Delta1925:
I have been using this patch for a few months now and I haven't noticed any issue with it. However I can't guarantee that this will work for everyone.
I think that #377 should probably be enough to justify the usage of the `unlock-dialog` permission and get the extension published.

aunetx:
Okay, well then I merge this! Thanks a lot :)

Issue #530, opened 2024-02-21
Broken compatibility with Burn My Windows on Wayland if fractional scaling is off


FelixFourcolor:
(continuation of #528)

Another consequence of commit e242657 (PR #522) is that when fractional scaling is off, [Burn My Windows](https://github.com/Schneegans/Burn-My-Windows) applies its affect at the wrong area.

Without e242657 (pre-v55), the effect is applied exactly on top of the window: [screencast](https://github.com/aunetx/blur-my-shell/assets/34260584/c0494ec9-3bd5-489f-8e80-55da774b013c)

With e242657 (v55), the effect overflows the window boundary: [screencast](https://github.com/aunetx/blur-my-shell/assets/34260584/9796e671-9fc6-4250-9dd7-2686a8abc068). The overflow part looks like the overflow blur box as described in #528, I assume it's not a coincidence.

I only show in the screencasts one of Burn My Windows' many effects, but in my testing I find most - if not all of them - are similarly broken.

After testing I find the same situation as in #528:

|| With e242657  | Without e242657 |
|- | ------------- | ------------- |
| Fractional scaling on | working  | broken (#362) |
| Fractional scaling off | broken (this issue) | working |

aunetx:
Working on this :)

aunetx:
Uh, actually not really, working on #528 but I guess once it's solved this bug will be too, as Burn my Windows simply takes the blur effect into account in its frame rectangle calculation

aunetx:
This should be fixed by #534, waiting for some testing :)

PR #534, opened 2024-02-24
Fix scale-monitor-framebuffer inconsistencies

### LINKED ISSUE #528: Broken application blur on Wayland if fractional scaling is off
~~v55, when fractional scaling is off, unfocused windows are incorrectly blurred~~

~~The window on the left is focused. On the unfocused windows, the entire thing is blurred instead of only the background.~~ 
(See EDIT below.)

Screenshot from 2024-02-19 02-11-08

I think commit e2426578f5651dce33616cb3ba22e4339f199cc7 is the culprit, because when I revert it the problem disappears. But the commit fixes #362. So basically the situation is:

|| With e2426578f5651dce33616cb3ba22e4339f199cc7  | Without e2426578f5651dce33616cb3ba22e4339f199cc7 |
|- | ------------- | ------------- |
| Fractional scaling on | working  | broken (#362) |
| Fractional scaling off | broken (this issue) | working |

----------------
EDIT: I had an incorrect understanding of the issue (thanks to @alesya-h for pointing it out). It's not inherently about blurring of unfocused windows, but about the size and placement of the blur box. This screenshot shows it more clearly.
Screenshot from 2024-02-20 17-05-24

If there are multiple windows on-screen, the unfocused ones will be put behind the blur box of the focused window, thus it looks like unfocused windows are incorrectly blurred.
Screenshot from 2024-02-20 17-05-58

aunetx:
Should fix #528 and #530

### DIFF
--- a/src/components/applications.js
+++ b/src/components/applications.js
@@ -6,6 +6,8 @@ import * as Main from 'resource:///org/gnome/shell/ui/main.js';
 import { PaintSignals } from '../effects/paint_signals.js';
 import { ApplicationsService } from '../dbus/services.js';
 
+const mutter_gsettings = new Gio.Settings({ schema: 'org.gnome.mutter' });
+
 
 export const ApplicationsBlur = class ApplicationsBlur {
     constructor(connections, settings, _) {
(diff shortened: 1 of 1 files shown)


FelixFourcolor:
Error from extension manager, extension cannot be enabled: `ReferenceError: Gio is not defined`.

FelixFourcolor:
You just missed one line `import Gio from 'gi://Gio'` in `src/components/applications.js` (before calling `const mutter_gsettings = new Gio.Settings({ schema: 'org.gnome.mutter' });`). Otherwise everything works.

aunetx:
Yes sorry, thank you very much for the testing!

PR #544, opened 2024-03-17
Handle the 'unlock-dialog' state

aunetx:
However, this makes the overview blur disappear right before sleep, which is not very great...


aunetx:
Well, it also happen if we do not take care of `unlock-dialog`, so no regression here at least

PR #546, opened 2024-03-23
Use `BackgroundManager` for static blur

aunetx:
This uses the native `BackgroundManager` class to replace the use of `BackgroundActor`, which prevents the need to refresh the image very often and is much more efficient.

However, this leads to two bugs that I need to fix:
- the image is clipped to where the user can see it (very efficient), but this means that the blur does not see past the frontiers of its blurred widget. For overview, animation and screenshot blur we don't care, but for panel blur (and dash-to-dock blur once it's implemented) it is quite important as it means the panel won't blend into the overview as it did before
- moving elements is quite slow, so when using hidetopbar extension, the panel being hidden/shown with a drop-down animation means having 2/3 fps... which is not good indeed, but at least does not affect the "normal" user

I may hence need to not use this technique for the panel blur, eventhough it would decrease quite a bit the interest of this change.


aunetx:
First point is fixed, it took me a good time for nearly nothing :)

aunetx:
Well, point 2 is visibly fixed too :)

aunetx:
Well that should fix a lot, good PR, I luv it

PR #556, opened 2024-03-27
Implement Monte Carlo blur

aunetx:
Just a proof-of-concept for the moment, there is no way to change the parameters other than in the code for example! Based on https://www.shadertoy.com/view/MdXXWr#

This probably won't be merged, because I intend to really beef up the extension before that:
- adding "pipelines" which would permit the users to combine different effects together
- permitting the user to choose which pipeline to use for which component
- implementing different effects (this one, the gaussian blur effect, etc)
- PROFIT hehe

However, I first need to fix a big performance issue, because the effects are currently being redrawn very often, which causes low FPS (especially when using big effects, for example augmenting the iterations for the Monte Carlo blur or the radius for the gaussian blur). So I need to find a way to redraw the effect way less often, and then I may do it!

PS you may play with the parameters in `src/conveniences/blur_creator.js`, lines 41 to 49. More iterations is slower and converges towards a gaussian blur, more radius is not slower but is not very pretty when there are not enough iterations, and `use_base_pixel` makes an interesting variation when using low iterations!

With 5 iterations, radius of 2 and `use_base_pixel` to `false`:
image

With 15 iterations, radius of 15 and `use_base_pixel` to `false`:
Capture d’écran du 2024-03-27 22-44-31

With 7 iterations, radius of 5 and `use_base_pixel` to `true`:
image

### DIFF
--- a/src/conveniences/blur_creator.js
+++ b/src/conveniences/blur_creator.js
@@ -3,6 +3,9 @@ import St from 'gi://St';
 import * as Main from 'resource:///org/gnome/shell/ui/main.js';
 import * as Background from 'resource:///org/gnome/shell/ui/background.js';
 
+import { BlurEffect } from '../effects/blur_effect.js';
+import { MonteCarloBlurEffect } from '../effects/monte_carlo_blur.js';
+
 export const create_background = function (
     monitor_index,
     background_managers,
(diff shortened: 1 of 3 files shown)


aunetx:
This is superseeded by #559, which implements monte carlo blur too -- but as a side-effect of the big changes that are coming :)

PR #559, opened 2024-03-29
Use pipelines instead of per-component parameters

aunetx:
This permits to change the way the preferences are constructed: the user is able to set pipelines, each within which is a list of effects that can be parametrized.

Then, for each component, the user can choose to use one pipeline or another. The goal is to:
- simplify the preferences if there are more effects that are added, it is quite simple to see how the "color" and "noise" effects are already a UX problem
- permit the addition of many effects without disorienting the user (as they would only appear in a drop-down in the preferences)
- simplify how the each effect is coupled to its actor, by coupling width, height, scale... automatically
- hopefully move away from a component-centric approach to be able to blur less easily accessible components (drop-downs, popups...). But that is really very unsure.


aunetx:
Quite OK now! Needs some testing :)

ryzendew:
image
seems like dock blur isn't working and the new blur style also

aunetx:
Yes I've not touched to the dock yet, and neither to non-static blur! I think that's the next thing on my list (just after fixing a little bug)

A what do you mean by the "new blur style"?

ryzendew:
> Yes I've not touched to the dock yet, and neither to non-static blur! I think that's the next thing on my list (just after fixing a little bug)
> 
> A what do you mean by the "new blur style"?

I think you called it Monte Carlo it was just a bunch of noise

aunetx:
There are now only four things left:
- adding the pipelines / blur mode chooser to Windows List (easy, quick to do, will implement tomorrow during class hehe)
- see what I do for application blur!!! I may keep it like that *for the moment* and merge this pull request, because it may take some more days... But in the end, I think the application will suffer the same treatment as the rest of the components, and you will be able to choose between static blur and dynamic blurring!
- see for the translations, I just need to change `effets.js` to something `gettext` can access
- see how I do the user preferences port... now that's not a funny thing to do, but I have an idea:
  - for the overview, try to use the previous config (native static gaussian blur, with the same preferences); else use `pipeline_default` as it is now
  - for the panel, if the user was using dynamic blur, keep it like this and if `customize` is activated, copy the sigma and brightness into the panel's ones; if he is using static blur, same thing as for the overview
  - for dash-to-dock, same thing as for the panel
  - for windows-list, same thing as for the overview
  - for the applications, probably keep dynamic blur with same settings as before, but maybe switch to static blur directly if it is more efficient? As the dynamic applications blur is quite slow IMHO

By the way, @a-parhom, if you have some time, I would love if you have some things to improve the performances!

I think I did everything I could to improve the perfs by reusing every effect possible, re-creating blurred actors way way less often than before (I think I am down from creating 1 blurred actor every 5 seconds to creating 1 every 5 minutes for a normal utilisation, and maybe down to a dozen actors per session if the user has only one monitor, even if the user changes the background!!!).

However, the `Clutter.ShaderEffect` we use are re-painted very often from what I can see (when adding a log in the `vfunc_paint_target` of either the gaussian blur or the monte carlo blur applied to the overview, I saw that it updated during every animation of the overview, and when I was simply hovering above it... which is clearly not normal, as the updates should happen only when the blurred widget actually change, which is very vary rare now). This causes very slow performances, especially when using the Monte Carlo blur with high iterations, or when using the non-native gaussian blur with a normal radius. So there are two possible things to do (and if possible, the two at the same time):
- reduce the size of the texture, the same way `Shell.BlurEffect` does not, but I have not really looked into it to be honest; and it should be optional as it would clearly make the effects less precise (especially the Monte Carlo blur which is less "blurry" than the gaussian blur)
- put the repainting of the effects in "pause" when the actor is not changing! But I don't have any idea how to do it, I asked some help upstream but had no answer; I will test and see if using the `Shell.GLSLEffect` changes anything (but I don't see the difference with `Clutter.ShaderEffect`); or else try again to get some advised help

By the way, quite weirdly, the native `Shell.BlurEffect` now supports stacking other effects above or below it, when using it in `ACTOR` mode! I don't know why, either it has been updated, or it is the way we were using the effects in the beginning that was problematic (with overflowing pixels notably).

aunetx:
image

With the new pixelize effect!

I honestly would really appreciate if some people could test this, because I won't release this new version without correct testing before... This is not finished, but still quite close to be!

ryzendew:
time for me to test it again

aunetx:
@ryzendew you can even test applications blur now!!!! I fixed the most buggy bug I ever encountered in this extension, so overall it should be A LOT more stable (and especially, not more annoying logs going everywhere)!!

aunetx:
Nearly finished!! Now needs only:
- do something about the translations, I get errors with `gettext cannot be called from outside the preferences` or something like that
- do something about the old preferences that works OK
- maybe (maybe?) try to have better prefs by default? Dash to Dock should obviously be using corners in static mode, ...

ryzendew:
> @ryzendew you can even test applications blur now!!!! I fixed the most buggy bug I ever encountered in this extension, so overall it should be A LOT more stable (and especially, not more annoying logs going everywhere)!!

will do

ryzendew:
ya application blur is alot more stable

aunetx:
Translations are done, now we only miss old preferences porting... I don't know if I really want to do it :/

a-parhom:
@aunetx I'm sorry, I didn't have any time these days, so I wasn't able to go further in texture downscaling. I hope, the next week will be not so hard and I will return to this. I'll also try to figure out how to stop unnecessary repaintings.

aunetx:
@a-parhom don't worry at all! This is really not so important, and to be honest I could do it myself too! The effects still have quite OK performances, and after all we can't do what's impossible :)

I think I will merge this PR now, as it is quite complete (although not perfect), and release the new version right after it; I've been testing it for quite some time now and even though it probably suffers bugs, it is still less buggy than the master branch right now.

PR #564, opened 2024-04-03
Fix transparent styling not applying to system actions in overview search

DaPigGuy:
Before:

image

After:

image


aunetx:
Thanks a lot, this looks better now!

PR #575, opened 2024-04-19
Maybe some hack to fix multiple monitors???

aunetx:


ryzendew:
has this been tested? if not i will try it

mendres82:
I just tested it. For me, the patch made the blur effect of the background in overview even more unreliable. Using GNOME 46.3 in a Wayland session on an AMD 6700XT with two displays connected, the first one with a resolution of 2560x1440 at 165Hz and the second one with 3440x1440 at 60Hz.

PR #582, opened 2024-04-22
Add a correct pixelize effect, HSL colorspace and fix crashes

aunetx:
Provides as-correct-as-possible downsampling and upsampling effects. This is useful for:
- the pixelize effect, which is effectively just downscale-rescale, but now it can be down with different averaging mode without destroying the performances
- every other effect!!! Because now we can downsample the image, apply the effect (provided we adapt a tiny bit each effect to act in a correct way), and upsample again (or not, if algorithms actually upsample themselves). This is still quite a bit of work, but I see the light now!

So all-in-one, this should bring more performances, more effects, more pixel-perfection, while still being cool.


aunetx:
@a-parhom sorry to ping you again, that was just to tell you I somehow managed to downscale images (well, I downscale in the top-left corner of the framebuffer, but still that's cool). So there is no need anymore to search a way to mimic how gnome-shell does it, as we have our own way now!

aunetx:
A bit stuck on what is left to do before merging, but I guess:
- remove "advanced" effects from the "new effects" dialog (because I don't want to have ANY issue opened about "downscaling effect" making the image appear in the corner of the overview...)
	- these effects are upscaling, downscaling, rgb-to-hsl and hsl-to-rgb conversions, and probably non-native gaussian effect too (which is too slow to be really usable I guess)
	- maybe put it into an alternative linked drop-down button "advanced effects"?
	- or rework the "new effect" dialog, at this point maybe I should be using multiple tabs but I really don't want to...
- maybe work on some effects to work well with downscaling/upscaling, and then provide a little drop-down button in the effect for the user to select a downscaling factor?
- maybe change a bit some effects so that they are only applied to some channels, but that's not a wanted feature for the moment so that could wait
- maybe change a bit the Monte Carlo blur to have a better (gaussian rather than uniform) noise function, because it is sometimes really ugly
- tweak the different effects for them to be *pixel perfect*, thanks to the heavy research I made in order not to have a single pixel out of the GLSL canvas (I probably should do something else of my life)
- remove all the branching in GLSL code (I don't know how to produce a single line of performant gpu code so probably I will steal someone else's brain to do it)

a-parhom:
> @a-parhom sorry to ping you again, that was just to tell you I somehow managed to downscale images (well, I downscale in the top-left corner of the framebuffer, but still that's cool). So there is no need anymore to search a way to mimic how gnome-shell does it, as we have our own way now!

That's good news! Anyway, I didn't have any time to look into it properly these days. Thank you for your hard work on this extension!

Issue #585, opened 2024-04-28
3 finger swipe left/right doesn't work after resume on Gnome 46 when blur-my-shell is enabled.


spxak1:
I have troubleshoot this to some extent but admittedly I have a large number of extensions and while it's certain it is an extension that causes this issue, by elimination troubleshooting method may not be conclusive.

So any way to actually monitor blur-my-shell by checking logs or otherwise, would be welcome.

The problem is very particular: 
* Start F40 and log in with all extensions enabled
* Try 3 finger swipes: up/down for overview and apps and back, left/right to move workspaces
* Suspend (to RAM)
* Wake up
* Try 3 finger swipes: up/down works fine, left/right doesn't move workspaces anymore

Now, disable ```blur-my-shell``` and repeat above sequence. Now left/right works.

As I said I have a large number of extensions so it may be a combination of blur-my-shell with another extension that causes it. I will troubleshoot further to see if such a combination exists and get back.

As for what is the actual problem, I cannot tell.

As far as I can tell mutter and gnome-shell appear to work fine after resume when the problem manifests.

```sudo libinput debug-events --verbose``` identifies the gesture as per normal with: 

~~~
event11  GESTURE_SWIPE_BEGIN     +29.925s	3
event11  GESTURE_SWIPE_UPDATE    +29.925s	3  1.57/-6.04 ( 1.58/-6.11 unaccelerated)
event11  GESTURE_SWIPE_UPDATE    +29.932s	3  1.79/-7.60 ( 1.81/-7.69 unaccelerated)
.
.
.
  ... event11 - gesture state GESTURE_STATE_SWIPE → GESTURE_EVENT_RESET → GESTURE_STATE_NONE
 event11  GESTURE_SWIPE_END       +30.096s	3
~~~

It's just that the workspaces won't move left/right. 

This is of course on Wayland.

Thank you.

l3ifk:
I can confirm this behavior and the addition that it is not just wayland but X11 too.

flexagoon:
I identified the cause of the issue:

> Both [extensions] patch the WorkspaceAnimationController.prototype._prepareWorkspaceSwitch function.

> [Blur My Shell] doesn't actually get disabled, it only disables user session mode.

> What seems to happen is that the `_disable_user_session()` function also restores the patched method, but since it's not the actual `disable()` function, extensions don't get rebased.

See https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/7629

[...]

PR #588, opened 2024-05-03
fix: don't restore patched prototypes when suspending

flexagoon:
Patched prototypes should only be restored in the `disable()` method of the extension, since gnome enforces extension rebasing when disabling them. Restoring protos in any other place may lead to infinite recursion if a different extension is also patching the same function.

See https://gitlab.gnome.org/ewlsh/gjs-guide/-/issues/75, https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/7629

Fixes: #585

### DIFF
--- a/src/components/overview.js
+++ b/src/components/overview.js
@@ -27,6 +27,7 @@ export const OverviewBlur = class OverviewBlur {
             { name: 'bms-animation-backgroundgroup' }
         );
         this.enabled = false;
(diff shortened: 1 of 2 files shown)


aunetx:
Thanks a lot! I would not have found it :)

flexagoon:
@aunetx when will the next version be released? This is a blocking issue for rounded window corners.

PR #619, opened 2024-07-04
Enable blur for floating panel (static blur) and fix for panel_box dynamic blur

neuromorph:
Hello Aurélien,

I tried BMS for floating panel since BMS now supports rounded corners (static) but faced an issue. This is a small PR to fix it. Let me explain the scenario below so you can see if you would like to go this way.

**Current state for floating panel with Static blur**:

Blur is applied to panel-box and so goes beyond the panel -
current_static

Current state for floating panel with Dynamic blur:
Code uses x, y for the Panel but Width, Height for the Panel-Box causing this mismatch -
current_dynamic

**How it would look with the PR**:

Floating panel with Static blur (blur applied to Panel):
new_static

Floating panel with Dynamic blur (blur applied to Panel-Box since rounded corners not supported in Dynamic):
Not floating anymore since blur applied to Panel-Box. We can apply it to Panel instead but it will be a floating rectangle without the rounded corners. So instead this could be treated like no-support for floating with Dynamic (yet).
new_dynamic

Some more variations:
Regular panel:
v1

Regular with colored button islands:
v2

Floating with colored button islands:
v3

Note: The panel effects for Floating or Colors are added using [Open Bar](https://github.com/neuromorph/openbar) extension.

### DIFF
--- a/src/components/panel.js
+++ b/src/components/panel.js
@@ -237,16 +237,14 @@ export const PanelBlur = class PanelBlur {
             // sometimes yields NaN (probably when the actor is not fully
             // positionned yet)
             let [p_x, p_y] = panel_box.get_position();
-            let [p_p_x, p_p_y] = panel_box.get_parent().get_position();
-            let x = p_x + p_p_x - monitor.x;
-            let y = p_y + p_p_y - monitor.y;
-
-            background.set_clip(x, y, width, height);
-            background.x = -x;
-            background.y = -y;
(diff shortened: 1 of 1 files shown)


aunetx:
Hello @neuromorph, thank you very much for this change! It seems to work both on vanilla GNOME and with dash-to-panel so that's good for me :)

Merging this right now, and this will be included in next release!

PR #626, opened 2024-07-23
Use meson build system

carlwgeorge:
This standardizes a system-wide installation, which is useful for distros who create native system packages of this extension.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,17 @@
+project('blur-my-shell')
+
+gnome = import('gnome')
+i18n = import('i18n')
+
+uuid = 'blur-my-shell@aunetx'
+rdnn = 'org.gnome.shell.extensions.blur-my-shell'
+
+datadir = get_option('datadir')
+extension_dir = datadir / 'gnome-shell' / 'extensions' / uuid
+
+subdir('src')
+subdir('resources')
+subdir('schemas')
+subdir('po')
+
+install_data('metadata.json', install_dir : extension_dir)
--- a/po/meson.build
+++ b/po/meson.build
@@ -1,0 +1,1 @@
+i18n.gettext(uuid, preset : 'glib')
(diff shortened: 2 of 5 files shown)


aunetx:
Hello, and thank you for your work!

I don't know how I feel about using Meson myself, to be honest after 3 years using it from time to time I still have no clue about how it really works... But that could be a step in the right direction!

Would it be good to keep my Makefile along with Meson, or is it bad practise? I like it a lot to be honest (it does exactly what I want, and I don't want to have to remember gettext commands); but the build/install step is not the same as Meson... Maybe it could be possible to have Make calling Meson so that it changes nothing for me, but everything stays in sync?

In any way, I will merge this after the other pending PR :)

carlwgeorge:
Hello, and you're quite welcome.  I'm fairly new to meson myself, and part of submitting this PR involved learning more about it.

I think make might be a bit more flexible with arbitrary custom targets (such as your `test-shell` and `test-prefs` targets), where as meson focuses heavily on standard "targets" as subcommands (setup, compile, install, and test).  From what I've learned so far, [a make target can be a file name or an action to carry out](https://www.gnu.org/software/make/manual/html_node/Rule-Introduction.html), but [in meson it's just a file name](https://mesonbuild.com/Reference-manual_returned_build_tgt.html#build-target-build_tgt-extends-tgt).  I don't know what the meson equivalent of a make action target would be.  Like I said, I'm still learning.

I've seen some projects use multiple build systems, such as make and meson.  Perhaps the best approach is to use meson for the standard action targets, and use make for custom ones.  If you agree I can adjust this PR to remove the `build` and `install` targets from the Makefile to avoid duplication.  Or we could have those make targets call the equivalent meson commands as you suggested, which seems a bit strange to me but is probably fine.  Another choice would be to use meson for the standard stuff, convert the custom targets into simple bash scripts, and ditch make entirely.  Lots of ways to go about it, and I can work with whatever you prefer.

PR #630, opened 2024-07-29
Fix app folder blur animation disregarding shell settings

DaPigGuy:
This is most noticeable in a VM where animations are disabled. You can test this by disabling animations and opening an app folder in the app grid 

```
gsettings set org.gnome.desktop.interface enable-animations false
```


aunetx:
Hello again, and thanks a lot! This is perfect, merging this right now

Issue #637, opened 2024-08-16
broken blur on panel


Nekothegamer:
updated the extension to the most recent version and got greeted with this broken blur on the panel 

![Schermata del 2024-08-16 04-31-25](https://github.com/user-attachments/assets/08fa7fd7-0fa5-4d8c-acf3-c51145294733)

mvolf:
Having the exact same issue as @Nekothegamer .

![Screenshot from 2024-08-16 07-59-55](https://github.com/user-attachments/assets/042a018b-de47-41e5-a62e-4c52756bbf1a)

azrdn:
+1
I believe it has something to do with incorrect blur location when the external monitor set as a main monitor.

<details>
<summary>Dynamic Panel Blur</summary>
Main Monitor on the right side

![Screenshot from 2024-08-16 11-36-55](https://github.com/user-attachments/assets/894d63d6-956f-41f7-8133-e44f53757e40)

Main Monitor on left side
![Screenshot from 2024-08-16 11-37-31](https://github.com/user-attachments/assets/28e91fe5-c61c-44a5-948d-375519479b77)

Main Monitor on bottom side
![Screenshot from 2024-08-16 11-37-58](https://github.com/user-attachments/assets/2419e027-9bed-4fd5-9be5-99e83b2d7c9e)

<br>
</details>

My temporary workaround is to set display as "External Only", then set it back to "Join Displays"

corngoblin:
I'm also having this issue
Fedora 40 Gnome 46

aunetx:
Hello, sorry for the bugs, I would love to do something but I still don't have another monitor lying around to test it... (I though I had a shitty little 1024×768 monitor where I am now, but I don't find it...).

It somebody can investigate I would be glad to help but else it will need to wait a little bit :/

TheRsKing:
![grafik](https://github.com/user-attachments/assets/ec8cdf30-635b-4041-a6a8-cd6d1084c9de)
yeah same issue
Aeon Desktop
gnome 46

kimjongbing:
Having the exact same issue 
Gnome 46
Arch

aunetx:
Maybe you can try reversing the changes done in #619 and telling me if it works again?

Nekothegamer:
uhm... how do i do that?

Nekothegamer:
as in: how do i revert those changes? i should use the version prior to this release?

corngoblin:
[...]

PR #638, opened 2024-08-17
fix: multi-monitor panel blur - #637

neuromorph:
Hello Aurélien,

Added the fix for #637 as discussed. 

Thanks for quick update!


aunetx:
Perfect, thank you very much!

PR #679, opened 2024-10-18
Added missing German translations

Vl4dimyr:
I added all the missing German translations.

German words tend to be longer than English words, so the `Gaußscher Weichzeichner (fortgeschrittener Effekt)` might be too long for the line it's supposed to be in.

If that's the case, I can change `(fortgeschrittener Effekt)` to `(fortgeschritten)`  in all translations that use this phrase to ensure the text fits and remains consistent.

The rest of the translations should be fine in terms of length.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -71,7 +71,7 @@ msgstr ""
 
 #: src/effects/effects.js:86
 msgid "Gaussian blur (advanced effect)"
-msgstr ""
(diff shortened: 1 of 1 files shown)


aunetx:
Thanks a lot! I will merge this as-it, and if you (or anybody else) want to modify it to keep consistency that can be done later too :) Thanks!

PR #717, opened 2025-02-06
Update the panel font when updating its background

gautier-ponsinet:
When the gnome color-scheme is set to 'prefer-light', and the options 'force-light-text' and 'Disable when window is near' are enabled, the text of the panel would appear white on white and thus be unreadable if a window is near the panel. This commit fixes this issue by updating the text color of the panel whenever the panel background is updated, and by forcing light text only when the panel's background is blured/overrided (and the same settings as above are applied).


aunetx:
Thanks a lot for the fix! Merging this, will be included in next release :)

PR #750, opened 2025-05-12
Fixed transparency logic for just perfection bottom panels

HansHT:
### Summary
Fixes an issue with the _"disable when a window is near"_ logic for users using Just Perfections bottom panel setting.

### Changes

- Detects whether the panel is at the top or bottom of the screen
- Adjusts window overlap logic to account for bottom-aligned panels
- Keeps the original **5 * scale** margin

### Testing

- Gnome 48 (Wayland)
- Just Perfection: Tested top and bottom panel alignments.


aunetx (approved):
LGTM, thanks a lot for the change!

PR #771, opened 2025-08-14
Fix race condition with Dash to Panel on startup

Anduin2017:
Hi, this PR fixes a long-standing race condition that occurs when enabling the extension while Dash to Panel (DTP) is already active.

The Problem:

On startup, blur_dtp_panels() is called before DTP's panel actors have been fully allocated and drawn by the compositor. This results in the blur effect being created with incorrect dimensions (e.g., zero height), leading to a broken visual appearance. The issue resolves itself if the panel blur is toggled manually later, as the UI is stable by then.

Repro:

* Install both dash to panel and blur my shell.

```bash
anduin@anduin-lunar:/usr/share/gnome-shell/extensions/blur-my-shell@aunetx$ gnome-extensions disable blur-my-shell@aunetx
anduin@anduin-lunar:/usr/share/gnome-shell/extensions/blur-my-shell@aunetx$ gnome-extensions enable blur-my-shell@aunetx
```

image

Mitigation:

```bash
anduin@anduin-lunar:/usr/share/gnome-shell/extensions/blur-my-shell@aunetx$ dconf write  /org/gnome/shell/extensions/blur-my-shell/panel/blur false
anduin@anduin-lunar:/usr/share/gnome-shell/extensions/blur-my-shell@aunetx$ dconf write  /org/gnome/shell/extensions/blur-my-shell/panel/blur true
```

image

The Solution:

This fix wraps the core logic of blur_dtp_panels() inside a GLib.idle_add() call. This defers the execution to the next idle cycle, giving the compositor enough time to calculate the final geometry of the panel actors.

This approach is minimally invasive as it doesn't change any of the calling logic in the enable() method, but ensures that the blurring operation always runs against a fully-rendered panel.

This should robustly fix the compatibility issue with Dash to Panel.

### DIFF
--- a/src/components/panel.js
+++ b/src/components/panel.js
@@ -1,4 +1,5 @@
 import St from 'gi://St';
+import GLib from 'gi://GLib';
 import Meta from 'gi://Meta';
 import * as Main from 'resource:///org/gnome/shell/ui/main.js';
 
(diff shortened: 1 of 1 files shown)


Anduin2017:
@aunetx review please?

PR #807, opened 2025-11-03
Added support to gnome 49 so it can work in the newer version

tsubaie:
One line to add support to Fedora 43 and other distros


Humorist2601:
This can be closed

Issue #785, opened 2025-09-08
No blured Top Bar if Second Monitor is on top of the first


Karto007:
My Setup contains two Monitors and the second Monitor is placed on top of the first, so in the Settings i have configured the second Monitor over the first one. With this configuration de function to "unblur" or darken the Top Bar, if a Window is near, dose not work properly anymore. In this configuration the Top Bar is allways dark even if there isn't any window open. I found out, that if I position the second Monitor in the Settings on the left, right or under the first Monitor, the function works as it should an darkens the Top Bar only if a Window is near.

mammo0:
I also faced this issue. The following hackish workaround fixed it for me. But I don't know if this has any negative side effects.

```patch
--- src/components/panel.js
+++ src/components/panel.js
@@ -88,6 +88,16 @@
         // This is crucial to ensure the panel actors have been allocated their
         // final size and position by the compositor, avoiding race conditions
         // during extension startup.
+        // if main panel is not included in the previous panels, blur it
+        if (
+            !global.dashToPanel.panels
+                .map(p => p.panel)
+                .includes(Main.panel)
+            &&
+            this.settings.dash_to_panel.BLUR_ORIGINAL_PANEL
+        )
+            this.maybe_blur_panel(Main.panel);
+        
         GLib.idle_add(GLib.PRIORITY_DEFAULT_IDLE, () => {
             if (!global.dashToPanel?.panels) {
                 return GLib.SOURCE_REMOVE;
@@ -101,14 +111,14 @@
             });
     
             // if main panel is not included in the previous panels, blur it
-            if (
-                !global.dashToPanel.panels
-                    .map(p => p.panel)
-                    .includes(Main.panel)
-                &&
-                this.settings.dash_to_panel.BLUR_ORIGINAL_PANEL
-            )
-                this.maybe_blur_panel(Main.panel);
+            //if (
+            //    !global.dashToPanel.panels
+            //       .map(p => p.panel)
+            //        .includes(Main.panel)
+            //    &&
+            //    this.settings.dash_to_panel.BLUR_ORIGINAL_PANEL
+            //)
+            //    this.maybe_blur_panel(Main.panel);
     
             return GLib.SOURCE_REMOVE;
         });
```

mammo0:
Created PR #808 which solves this issue for me.

Feel free to test it. Thank you.

mammo0:
Hello @arthur-robog, thanks for your findings.

[...]

PR #808, opened 2025-11-04
Fix blurring of original panel when using DTP

mammo0:
Hello,

without Dash to Panel (DTP) extension the blurring of the original Gnome panel is done outside of the `GLib.idle_add()` context, which works without any issues. With the current implementation and enabled DTP #785 occurs.

I think the blurring of the DTP panels works just fine in the `GLib.idle_add()` context but not for the original panel. Therefore, I moved the code which blurs the original panel out of this context. It is now handled in the `blur_existing_panels()` method like it is done if the DTP extension is not present.

This fixes the above issue for me.

### DIFF
--- a/src/components/panel.js
+++ b/src/components/panel.js
@@ -77,6 +77,10 @@ export const PanelBlur = class PanelBlur {
             // blur already existing ones
             if (global.dashToPanel.panels)
                 this.blur_dtp_panels();
+
+            // blur main panel if requested in settings
+            if (this.settings.dash_to_panel.BLUR_ORIGINAL_PANEL)
+                this.maybe_blur_panel(Main.panel);
         } else {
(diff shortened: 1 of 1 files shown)


Neo-29:
If you could fix the blurred in overview too, it would be wonderful

mammo0:
First of all it your bug is in connection with another extension. I do not use this one, so I'm currently not affected by this bug.

And as stated out in your linked issue it's an upstream bug. So it must be fixed by Gnome developers.

mammo0:
I do not use the "Rounded Window Corners" extension. And I'm not affected by this bug. And I'm also not a Gnome developer. So I can't help here.

aunetx:
Hello, thanks a lot for this PR and sorry for not reviewing it before. It looks perfect to me and simplifies the logic a bit!

Issue #810, opened 2025-11-13
[feature] Luminosity-preserving window blur


fractioneater:
A blur is nice for making windows legible over the desktop background.

<img width="1254" height="865" alt="Image" src="https://github.com/user-attachments/assets/ba555905-e931-4be5-b11e-9d387015e30f" />

But when you have a light wallpaper or the window is on top of light window, it's harder to read.

<img width="1254" height="865" alt="Image" src="https://github.com/user-attachments/assets/966b9220-83e8-4c5a-a6d0-0529deb5e94b" />

My current solution for this is to reduce the brightness of the application blur by a lot. Now I can read it perfectly fine over a white background, but it's unnecessarily dark over a dark background (the normal circumstance).

~~The nicest-looking solution, in my opinion, is to implement a [color burn effect](https://www.ctnicholas.dev/articles/which-blend-mode#color-burn), which darkens the background image to the color of the source. For example, a dark background would stay dark (not getting any darker) and a light background would become dark.~~
**Later me knows better! Color burn isn't quite right for this.**

I'm willing to submit a PR if I get confirmation that it's possible (if the shader files can pull the texture from the window separately from the background texture).

fractioneater:
I've made [a mockup in Figma](https://www.figma.com/community/file/1571326466134938125/window-darkening) of the effect. As you can see, even on a solid white background, it will be legible without over-darkening the rest of the image.

fractioneater:
!!!

<img width="1738" height="1126" alt="Image" src="https://github.com/user-attachments/assets/0fe90ecc-236e-4098-9665-6ba5ac823f7c" />

Issue #762, opened 2025-06-24
[feature] static application blur


Inderix:
I think it'd be neat to be able to set a static blur for apps like you can for dash-to-dock and the top panel blur, retaining the "look" of background blur in apps, but with better performance and the ability to use the rounded corner pipeline to fix the corner artifacts.

It'd also allow for replicating Microsoft's "Mica" material used in Fluent design apps and Windows 11:
https://learn.microsoft.com/en-us/windows/apps/design/style/mica

PR #819, opened 2025-12-26
Mica effect (static application blur)

### LINKED ISSUE #223: Feature Request: Material Mica for Applications
Windows 11 has a nice effect for blurred apps: instead of blurring the contents behind, it only blurs the wallpaper and uses that as background.
This makes the apps have the colors of your wallpaper and increases performance, which is nice.

Would something like this be possible with this extension?

Screenshot of mica effect

fractioneater:
Fixes #223, #810, #811, and #762, most of which are the same thing.

### Static application blur
The applications prefs page now has a static/dynamic picker like panel and dash-to-dock. A pipeline chooser is visible when static blur is selected, and the dynamic blur options are available when dynamic is selected.

Screenshot From 2025-12-26 17-48-01

image

_(the transparency is my GTK theme's fault)_

### Things I've tested

- Editing the settings will update windows dynamically, even switching static to dynamic and vice versa
- Window size, visibility, and position changes are handled as expected

### Why this is still a draft

- [ ] **Wayland fractional scaling** `applications.js, line 179`  
  I tried with and without dividing by the scale variable, and it didn't appear to make a difference. Has Wayland changed since it was implemented, or is something weird on my end?
- [X] **Dynamic enable/disable** `applications.js, line 422`  
  ~~Is there a less brutal way to redo all blur effects than `disable()` and `enable()`-ing the whole thing?~~
  Resolved—it now uses an already-existing function for pipeline changes.
- [X] **Whatever this is** `luminosity.glsl, line 32`  
  ~~I didn't intend to keep this in here. It'll go away and reappear in another PR—as its own effect—soon.~~
  Resolved. The other PR is #821.
- [x] **The other luminosity effect**  
  ~~I'm aware of #773, and this PR shadows the name "luminosity". I originally called my effect "mica" instead of luminosity, but mica is much more than this one effect; it mostly means a static blur.~~
  Resolved by removing luminosity effect in favor of color effect (from pull #821).
- [x] **X11 windows layering issue**  
  ~~Various X11 apps are covered by the mica layer.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,8 +47,8 @@ A GNOME Shell extension that adds a blur look to different parts of the GNOME Sh
       - similar to the Windows "Acrylic" effect
       - shows other windows behind the application
       - more comprehensively tested — less likely to cause problems
-    - you can select the opacity of the window that is above the blur: a lower opacity means it will be less legible
-    - but you can choose to make the focused window totally opaque so that you can enjoy your blur while always having a legible window you work on! (for this, use a transparent GTK theme)
+    - you can select the opacity of the window that is above the blur: a lower opacity means it will be less legible; and a totally opaque setting can be used in pair with a transparent GTK theme to make the background of the windows blurred without touching at its content
+    - but you can choose to make the focused window totally opaque so that you can enjoy your blur while always having a legible window you work on!
     - you can activate an option to better blur the windows while in overview, although it won't make the blur perfect
     - and there are two modes for blurring applications:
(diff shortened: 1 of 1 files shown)


fractioneater:
New discovery: this has some issues with fullscreen windows (noticed with OrcaSlicer).

a-lupow:
Wow, nice and quick fix, thank you. I will let you know tomorrow how it goes

PR #855, opened 2026-04-12
Add -y flag to dnf install command

phantomcortex:
This change is small but important.

If attempting to install `gnome-rounded-blur` via the script on Fedora causes this:

image

It will abort without fail. this is because `dnf` wants to confirm that you actually want these changes but running `dnf` in a script will auto-abort unless it has the `--assumeyes` /  `-y` flag.


aunetx:
Thanks a lot!

PR #856, opened 2026-04-12
fix copy-and-paste execution for  `gnome-rounded-blur` script

phantomcortex:
the '$' prevents copy-and-paste execution so this should fix that.


aunetx:
Thanks!

Issue #631, opened 2024-08-04
Dash to Dock blur floods logs with 'Can't update stage views actor'


litemotiv:
[Version] 62

Enabling the Dash to Dock blur causes recurring log entries when the dock comes in or out of view:

```
Aug 04 22:44:59 fedora gnome-shell[2241]: meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
Aug 04 22:44:59 fedora gnome-shell[2241]: Can't update stage views actor dashtodockContainer [Gjs_dash-to-dock_micxgx_gmail_com_docking_DashToDock] is on because it needs an allocation.
Aug 04 22:44:59 fedora gnome-shell[2241]: Can't update stage views actor unnamed [Gjs_dash-to-dock_micxgx_gmail_com_docking_DashSlideContainer] is on because it needs an allocation.
Aug 04 22:44:59 fedora gnome-shell[2241]: Can't update stage views actor dashtodockBox [StBoxLayout] is on because it needs an allocation.
Aug 04 22:44:59 fedora gnome-shell[2241]: Can't update stage views actor bms-dash-backgroundgroup [MetaBackgroundGroup] is on because it needs an allocation.
Aug 04 22:44:59 fedora gnome-shell[2241]: Can't update stage views actor bms-dash-blurred-widget [StWidget] is on because it needs an allocation.
Aug 04 22:44:59 fedora gnome-shell[2241]: Can't update stage views actor unnamed [Gjs_dash-to-dock_micxgx_gmail_com_docking_DashSlideContainer] is on because it needs an allocation.
Aug 04 22:44:59 fedora gnome-shell[2241]: Can't update stage views actor dashtodockBox [StBoxLayout] is on because it needs an allocation.
Aug 04 22:44:59 fedora gnome-shell[2241]: Can't update stage views actor bms-dash-backgroundgroup [MetaBackgroundGroup] is on because it needs an allocation.
Aug 04 22:44:59 fedora gnome-shell[2241]: Can't update stage views actor bms-dash-blurred-widget [StWidget] is on because it needs an allocation.
Aug 04 22:45:00 fedora gnome-shell[2241]: Can't update stage views actor dashtodockContainer [Gjs_dash-to-dock_micxgx_gmail_com_docking_DashToDock] is on because it needs an allocation.
[...]

PR #861, opened 2026-04-15
Fix Dash-to-Dock blur allocation race

ektorthebigbro:
## Summary
Fix a startup/runtime allocation race in the Dash-to-Dock integration.

## What changed
- defer blur setup until the Dash-to-Dock actors have a real allocation
- skip blur size updates while the dash container, dash, or dash background are still unallocated

## Why
On GNOME Shell 49 with Dash-to-Dock enabled, Blur my Shell can attach `bms-dash-backgroundgroup` and `bms-dash-blurred-widget` as soon as the dock actor is added to the stage, before Dash-to-Dock has completed allocation. That produces repeated warnings like:

- `Can't update stage views actor dashtodockBox [StBoxLayout] is on because it needs an allocation.`
- `Can't update stage views actor bms-dash-backgroundgroup [MetaBackgroundGroup] is on because it needs an allocation.`
- `Can't update stage views actor bms-dash-blurred-widget [StWidget] is on because it needs an allocation.`

Deferring setup until allocation is available stops that invalid state instead of repeatedly resizing or repainting unallocated actors.

## Validation
- VS Code JS diagnostics: no errors in `src/components/dash_to_dock.js`
- local session patch applied successfully to the installed extension
- reloaded `blur-my-shell@aunetx` after applying the fix

### DIFF
--- a/src/components/dash_to_dock.js
+++ b/src/components/dash_to_dock.js
@@ -118,6 +118,11 @@ class DashInfos {
     }
 
     update_size() {
+        if (!this.dash_blur._has_valid_allocation(this.dash_container) ||
+            !this.dash_blur._has_valid_allocation(this.dash) ||
+            !this.dash_blur._has_valid_allocation(this.dash_background))
+            return;
+
         if (this.dash_blur.is_static) {
             let [x, y] = this.get_dash_position(this.dash_container, this.dash_background);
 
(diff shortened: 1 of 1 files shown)


genrichh93-ui:
Tested this locally on CachyOS GNOME with Dash to Dock and Blur my Shell, with Dash to Dock blur and “Override background” enabled.

After applying this PR locally and restarting the session, the warning spam appears to be gone and the dock visibility behavior works correctly again.

ChoeHa-U:
> Tested this locally on CachyOS GNOME with Dash to Dock and Blur my Shell, with Dash to Dock blur and “Override background” enabled.
> 
> After applying this PR locally and restarting the session, the warning spam appears to be gone and the dock visibility behavior works correctly again.

how to do, what excatly you did?

krissedout (approved):

krissedout:
lgtm! merging this, likely will help with #631

PR #862, opened 2026-04-17
gnome-rounded-blur script update

MeguMario:
- Fixed uninstallation on Fedora and Ubuntu because they both install the library on two different directory (`/usr/lib64` on Fedora and `/usr/lib/x86_64-linux-gnu` on Ubuntu)
- Now check for enviroment and install `git` if not installed (and `mutter` on ubuntu because they don't have that installed for some reason) before doing anything else.
- `-y` arg for `apt` as well

### DIFF
--- a/scripts/GUIDE.md
+++ b/scripts/GUIDE.md
@@ -28,11 +28,11 @@ You can get this library by getting it on [the AUR](https://aur.archlinux.org/pa
 
 - For `paru` users
 ```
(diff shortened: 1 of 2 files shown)


aunetx:
Hi, thanks for the update! Just changing the guide a bit (from #856, and so that it is the same as in the readme script commands), and merging :)

aunetx:
nvm, did not read the changes correctly, it's all good I will merge it directly

PR #863, opened 2026-04-17
Use adw-external-link-symbolic instead of just external-link-symbolic

MeguMario:
This fix the external link icon for those who do not use Yaru 
image

and Yaru default this to their icon pack anyway so no problem there


aunetx:
Yes thanks, i forgot about that!

PR #867, opened 2026-04-27
Fix: Secondary-monitor windows hidden after workspace switch when applications.BLUR is enabled

### LINKED ISSUE #866: Secondary-monitor windows hidden after workspace switch when applications.BLUR is enabled
### Description
 When `applications.BLUR` is enabled and GNOME's `workspaces-only-on-primary` setting is `true` (the default), all windows on a secondary monitor become hidden after a workspace switch. The windows still exist (they appear in `Alt+Tab` and the overview) but their compositor actors are .hide()`-ed, so visually the secondary monitor goes blank.

The symptom is most noticeable with 3-finger touchpad swipes, but the underlying code path runs for keyboard switches too.

  ### Steps to reproduce

  1. Plug in an external monitor (any setup with two displays works).
  2. Confirm GNOME default: `gsettings get org.gnome.mutter workspaces-only-on-primary` → `true`.
  3. Enable `applications.BLUR` in the extension settings (whitelist or enable-all, doesn't matter).
  4. Open one or more application windows on the **secondary** monitor.
  5. Switch workspaces (3-finger swipe, `Ctrl+Alt+Arrow`, etc.).

  **Expected:** Windows on the secondary monitor stay visible (they don't belong to a single workspace under workspaces-only-on-primary).

  **Actual:** They disappear after the switch finishes.

  ### Root cause

  `src/components/overview.js`, the patched `WorkspaceAnimationController._finishWorkspaceSwitch`:

  ```js
  if (outer_this.settings.applications.BLUR)
      for (let i = 0; i < w_m.get_n_workspaces(); i++) {
          if (i != w_m.get_active_workspace_index())
              w_m.get_workspace_by_index(i)?.list_windows().forEach(
                  window => window.get_compositor_private().hide()
              );
      }
```
Under workspaces-only-on-primary = true, secondary-monitor windows are flagged "on all workspaces" and are returned by list_windows() for every workspace. The loop iterates all inactive workspaces and hides every window in them — which hides those sticky windows globally.

## Proposed fix
Skip windows where meta_window.is_on_all_workspaces() returns true. Diff is one block in overview.js (will open a PR shortly).

**Environment**
1.  GNOME Shell 50.0 (Wayland)
1. Arch Linux, kernel 6.19.x
1. blur-my-shell version 71 (extensions.gnome.org) — same code path exists in current master
1. Hybrid NVIDIA RTX 3050 Ti / Intel
1. Mixed fractional scaling: laptop eDP-1 at 1.33x, external LG 4K at 1x (not required for repro, but my test setup)

**Workaround**
Disable applications.BLUR, or disable the extension entirely. Keyboard Ctrl+Alt+Arrow does not visibly trigger this in some setups (likely timing-dependent), but the same code path runs.

tghastings:
Fixes #866.

When `applications.BLUR` is enabled, the patched `_finishWorkspaceSwitch` iterates inactive workspaces and calls `.hide()` on every window returned by `list_windows()`. Windows on all workspaces — including secondary-monitor windows under `workspaces-only-on-primary = true` — appear in every workspace's list, so the loop hid them globally.

Skip them with `is_on_all_workspaces()`.

Tested on GNOME Shell 50 / Wayland with the repro from #866 — the bug is gone with the patch and reproducible without it.

### DIFF
--- a/src/components/overview.js
+++ b/src/components/overview.js
@@ -108,7 +108,13 @@ export const OverviewBlur = class OverviewBlur {
                     for (let i = 0; i < w_m.get_n_workspaces(); i++) {
                         if (i != w_m.get_active_workspace_index())
                             w_m.get_workspace_by_index(i)?.list_windows().forEach(
-                                window => window.get_compositor_private().hide()
+                                window => {
+                                    // skip windows on all workspaces (e.g. secondary
+                                    // monitor windows under `workspaces-only-on-primary`);
+                                    // hiding them on any inactive workspace hides them globally
+                                    if (window.is_on_all_workspaces()) return;
+                                    window.get_compositor_private().hide();
+                                }
                             );
                     }
 


aunetx:
Hi! Thanks a lot for the fix, merging this right now :)

Issue #844, opened 2026-03-25
Partial dash blur (patch code proposal)


opaczynski:
In the current version of Blur The Shell on Ubuntu 24.0.4 LTS, the dock looks like this in a dual-monitor setup:

<img width="78" height="1080" alt="Image" src="https://github.com/user-attachments/assets/481c0edc-865d-4db4-929f-c8b7c813bb68" />

From Ubuntu's perspective, it looks like this:
<img width="3000" height="1920" alt="Image" src="https://github.com/user-attachments/assets/92ccd381-5b90-4e90-8d88-6b15ceeb18da" />

---
That's why I wrote a small patch that _partially_ fixes the dock issue (I tested it on my secondary monitor, and the blur effect doesn't work there at all because the code calculates the blur position for the primary monitor, and I don't feel like updating the code that much). I adjusted the code to account for the offset, so the blur no longer cuts off halfway through.

Code in `components/dash_to_dock.js`:
 ```    
       get_dash_position(dash_container, dash_background) {
        var x, y;

        let monitor = Main.layoutManager.findMonitorForActor(dash_container);
        let offsetX = -monitor.x;
        let offserY = -monitor.y
        let dash_box = dash_container._slider.get_child();

        if (dash_container.get_style_class_name().includes("top")) {
            x = (offsetX + monitor.width - dash_background.width) / 2;
            y = offserY + dash_box.y;
        } else if (dash_container.get_style_class_name().includes("bottom")) {
            x = (offsetX + monitor.width - dash_background.width) / 2;
            y = offserY + monitor.height - dash_container.height;
        } else if (dash_container.get_style_class_name().includes("left")) {
            x = offsetX + dash_box.x;
            y = offserY + dash_container.y + (dash_container.height - dash_background.height) / 2 - dash_background.y;
        } else if (dash_container.get_style_class_name().includes("right")) {
            x = offsetX + monitor.width - dash_container.width;
            y = offserY + dash_container.y + (dash_container.height - dash_background.height) / 2 - dash_background.y;
        }

        return [x, y];
    }
```
Result:
<img width="91" height="1080" alt="Image" src="https://github.com/user-attachments/assets/25214426-1a6e-4b25-8e0a-2fb58d378cee" />

---
If any of you don't want to wait for the official update, you can simply overwrite function in the extension code and refresh the extension by typing “r” in ALT+F2 textbox.

opaczynski:
[...]

PR #883, opened 2026-05-14
Add blur support for GNOME Shell popups

### LINKED ISSUE #576: how to add blur to panel quick settings and other drop downs
i want to add blur here
---
Screenshot from 2024-04-19 23-24-59
---

krissedout:
This PR adds blur support for more GNOME Shell UI surfaces, including the calendar/notification popup, quick settings, notification banners, OSD popups, right-click menus, switchers, and dialogs.

image

The new **Popup blur** tab in preferences includes controls for blur, sigma, brightness, corner radius, and background override. The background override supports transparent, light, and dark styles, which helps avoid the double-tint look when blur is enabled.

image

For rounded corners with dynamic blur, popup blur uses the rounded blur helper when it is available, since native dynamic blur can handle the rounding in that case. If the helper is missing, the preferences UI now points users to it, and the shell blur path falls back to the existing corner effect where needed.

The popup blur implementation creates a non-reactive blur actor behind each detected shell surface and keeps it synced to the surface geometry. Updates and repaints are queued/coalesced, duplicate signal tracking is avoided, actor scanning stops once a suitable blur target is found, and hidden or unmapped shell actors are skipped.

I also renamed the **Hack level** preference to **Artifact handling**. The old name felt a bit internal/scary for a user-facing setting. The options are still the same, but the description now explains more directly that **No artifact** disables clipped redraws globally and can affect GNOME Shell performance. If this seems too out of scope I can undo it while keeping the other changes. Also I fixed a typo High Performances -> High Performance. 

Closes #576

Tested on:
- GNOME 50
- Wayland
- dynamic blur enabled
- rounded blur helper present


MeguMario:
If possible, I would like to see the corner radius being adjustable (or dynamically adjusted) for each of the popups (notifications, quick settings, right click menu,..) because all of them have different corner radius value while having the background set to anything other than Transparent (For example, I set the value to 35 which while matching QS and Notification section, does not match right click menu and notification popup )

image
image

krissedout:
fixed! it should now follow gnome's defaults for the corner radius however u can fine tune it in the settings, also added static blur support

image

MeguMario:
Nice! Though you should probably also override static blur radius settings (or present seperate radius option in pipeline settings) as well as it only has one radius settings for everything (same as dynamic blur before).

Also one minor nitpick I have is the way the popup disappear with blur isn't exactly smooth. Is it possible to have the blur being destroyed before the animation end?

krissedout:
i added a corner radius override to the static blur like for different parts, and i tried to make the destroy animations smoother by following gnomes opacity and stuff its not perfect but it should feel less choppy now

MeguMario:
Neat, thanks for the swift changes, really appreciate that!

While I was trying out the effect with static blur, somehow switching window made the entire screen flicker with the blur effect. Something might have gone wrong there.

https://github.com/user-attachments/assets/a94019bc-379b-4ddb-bd7b-eab10f7a3efd

Also this as well
```
May 16 23:02:34 meguro gnome-shell[28415]: Object Gjs_ui_altTab_WindowSwitcher (0x56226a898b30), has been already disposed — impossible to get any property from it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
                                           == Stack trace for context 0x5622694526e0 ==
                                           #0   7ffe9e716640 b   file:///usr/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell_blur_surface.js:408 (12345acccfb0 @ 36)
                                           #1   7ffe9e716700 b   file:///usr/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell_blur_surface.js:404 (12345acccf60 @ 26)
                                           #2   562269523010 i   file:///usr/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell_blur_surface.js:56 (12345accc470 @ 12)
                                           #3   562269522f88 i   file:///usr/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell_surface_fade.js:32 (12345accf5b0 @ 62)
                                           #4   7ffe9e7170e0 b   resource:///org/gnome/shell/ui/environment.js:71 (2565ae88dce0 @ 139)
                                           #5   7ffe9e7177a0 b   resource:///org/gnome/shell/ui/environment.js:187 (2565ae89a100 @ 45)
                                           #6   562269522ef8 i   resource:///org/gnome/shell/ui/init.js:20 (2565ae88d8d0 @ 48)
```

krissedout:
for the static blur flicker, shell static backgrounds were being created as full-monitor background actors first, then clipped afterward. in the window-switching path that could expose one frame before the clip was applied, which explains the full-screen blur flash.

for the disposed object warning, the fade completion callback could still query the switcher actor after GNOME had already destroyed it, i added root actor destroy handling and guarded those visibility/opacity reads so the callback should stop touching disposed actors. 

i also noticed the switcher blur was appearing before the switcher content, which made static blur show up as a blank blurred surface for a moment and fixed that as well!

aunetx:
Wow, thanks for this wonderful PR! I'll need some time to fully understand the code and merge it, but for sure this will be included in the next release :)

Without having looked at everything yet, i have some questions:
- I saw you created a lot of "helpers" for the shell blur (`shell_blur_surface.js`, `shell_message_stacks.js`, ...). If i understood correctly, this is a way to "automate" a bit more the handling of the effects for all those different kinds of popups. Do you think some of the components that were previously there could also use the same automation? For example, Coverflow Alt-Tab, Window List, or even the Dash/Panel (whose blur positioning with multiple monitors is so buggy that a more automated way of handling the blur could benefit them).
Also, the `src/components` directory was made for the "individual components", however you did put the helpers there. Do you think it would make sense to put them into `src/conveniences` instead, or make them have their own directory (or subdirectory of `src/components`), or just keep it as-it if you think it's better?
- There is not a lot of comments in the code. I'm not saying that commenting everything is better, but to be honest it makes me kind of lost... Do you think it would be possible to have some explanations for the important parts? Not when the code is self-explanatory of course, but you can look at `src/conveniences/pipelines.js` for example: I tried to explain the role of the module (even though it's a bit verbose...), and some parts of the code.
- You call blurring the popups "shell blur". It makes sense, but the extension is already called "Blur my Shell", so one would expect that the different options are for different parts of the shell.. maybe "popup blur" would be a more fitting name?

Sorry for the long message, but once again thank you very very much for this nice contribution!!

Proman4713:
This looks good! Looking forward to one day seeing it in Blur My Shell

aunetx:
Testing it a bit on GNOME 46, it works insanely well! Some comments:

- I have these logs appearing very regularly (with some different stack traces, i tried to find a variety of stacks):
  ```
  Object St.Button (0x61a81b44ca00), has been already disposed — impossible to access it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
                                               == Stack trace for context 0x61a818e9c670 ==
                                               #0   7ffedb767620 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:250 (c4d1210eba0 @ 60)
                                               #1   7ffedb7676e0 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:244 (c4d1210eb50 @ 20)
                                               #2   7ffedb7677e0 b   self-hosted:137 (16e7db4cd880 @ 245)
                                               #3   7ffedb767890 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:244 (c4d1210eb00 @ 31)
                                               #4   7ffedb767970 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:217 (c4d1210e9c0 @ 429)
                                               #5   7ffedb767a20 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:113 (c4d1210e650 @ 201)
                                               #6   61a818f63c68 i   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:130 (c4d1210e740 @ 64)
                                               #7   61a818f63bd8 i   resource:///org/gnome/shell/ui/init.js:21 (1db39a970bf0 @ 48)
  
  Object St.Button (0x61a81e31ced0), has been already disposed — impossible to access it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
                                               == Stack trace for context 0x61a818e9c670 ==
                                               #0   7ffedb767a10 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:225 (c4d1210ea10 @ 50)
                                               #1   61a818f63d88 i   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:213 (c4d1210e9c0 @ 374)
                                               #2   61a818f63cf8 i   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:113 (c4d1210e650 @ 201)
                                               #3   61a818f63c68 i   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:130 (c4d1210e740 @ 64)
                                               #4   61a818f63bd8 i   resource:///org/gnome/shell/ui/init.js:21 (1db39a970bf0 @ 48)
  
  Object St.Button (0x61a81b44ca00), has been already disposed — impossible to access it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
                                               == Stack trace for context 0x61a818e9c670 ==
                                               #0   7ffedb767930 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell_message_stacks.js:76 (c4d121196f0 @ 371)
                                               #1   7ffedb767a20 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:107 (c4d1210e650 @ 60)
                                               #2   61a818f63c68 i   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:130 (c4d1210e740 @ 64)
                                               #3   61a818f63bd8 i   resource:///org/gnome/shell/ui/init.js:21 (1db39a970bf0 @ 48)
  
  Object St.Button (0x61a821fe7090), has been already disposed — impossible to access it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
                                               == Stack trace for context 0x61a818e9c670 ==
                                               #0   7ffedb767850 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:250 (c4d1210eba0 @ 60)
                                               #1   7ffedb767910 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:50 (c4d1210e3d0 @ 18)
                                               #2   7ffedb7679e0 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell_message_stacks.js:70 (c4d121196f0 @ 239)
                                               #3   61a818f63cf8 i   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:107 (c4d1210e650 @ 60)
                                               #4   61a818f63c68 i   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:130 (c4d1210e740 @ 64)
                                               #5   61a818f63bd8 i   resource:///org/gnome/shell/ui/init.js:21 (1db39a970bf0 @ 48)
  ```
  Although it does not cause any crash for the moment.

- This is not very well integrated with the theme:
  - when I use dark mode, the buttons in the calendar / quick settings etc are very dark (using yaru, but it seems to be the same for the default adwaita theme); which does not play very nice with the transparency. Maybe the "override background" could actually provide transparency for such buttons, like I did for the appgrid folders for example (where not only the background but also some elements were changed)? image

  - with light theme, it becomes obviously a bit nasty: image
    At this point the solution may not be to override the light theme too much (this is what I did for the panel for example, but simply changing what the "light" option for background overriding means to have it indeed lighten the background instead of "lightly" darkening it (this was a very bad naming scheme of mine). With such a tweak (should be doable in CSS I guess, but for this particular screenshot emulated with a "color" effect with `rgba(1.,1.,1.,0.75)` under a gaussian blur), it would give something like this: image
    Which, along with some transparency tweaks for the buttons, would become much more attractive!
  - And the best would be to have an option to have an "automatic" option for the background styling, which would choose either "light" or "dark" depending on the shell theme, so that users activating the popups blur would not having something broken at first if the wrong theme is used.

- when using this in static mode with a pipeline that has no corner effect, well... no corners appear, which is a bit weird as they are still configurable. To correct this, the simplest way would be to have (in every pipeline object created) a hidden, non-removable "corner" effect with a specific name and with 0 corner radius by default, but whose radius would be changed dynamically when the user configures the corners in the preferences.
  The same method should ideally be applied to all other components I think, because at some point having corners should not really be considered an "effect" in itself (even if it is a GLSL effect), but more a configuration of the pipeline: I relied on the same kind of hacks as you did for the "enable corner rounding on maximized an fullscreen" option in Application blur, but this would prevent such a hack and allow the user to directly see a "Corners" option in the preferences for the Dash, Panel, Applications, Popups, ... without having to dig into the pipelines, when they don't have to do it for dynamic blur (with the library installed).
  To be honest this particular point is more for me, and this PR could be merged without fixing the corners issue: I will fix it myself just after merging, and have some mechanism I like!

MeguMario:
I just want to add that the way notification stack handling could be better too, as using this with a light theme result in this
image

On dark mode it's better for obvious reason
image

Both have broken shadow so I assume this need to be dealt with.

krissedout:
Thanks for the detailed review and testing!

For the disposed-object logs, I made the popup actor scanning more defensive. The scanner now tracks actors it has seen, marks them when their destroy signal fires, avoids queued rescans for destroyed actors, and wraps the recursive child/style lookups more carefully. I also applied the same kind of protection to notification stack handling.

For the theme integration, I adjusted the popup background override CSS quite a bit. The light style now actually lightens the popup surface instead of lightly darkening it. I also reworked the nested styling so the calendar/date menu keeps GNOME’s normal visual structure instead of adding filled backgrounds everywhere, while still restoring the current-day/selected-day highlight. Quick Settings also keeps checked toggle highlighting now, and normal quick toggle/menu surfaces use translucent backgrounds that fit better with blur.

For notifications, I removed the stacked-notification shadows that were getting clipped by the message list, added clearer stacked backgrounds, and fixed some light-mode contrast issues in notification headers/buttons.

I also moved more of the tint handling into the popup blur surface path so the blur/tint actors stay aligned better, especially for static blur. One remaining rough edge is that the background tint still is not perfectly synchronized with GNOME’s close animation in every path; in a few cases the tint can disappear slightly before the blur surface finishes fading. I don’t have a clean fix for that yet, but I wanted to call it out.

I also added an Automatic background style, which chooses light/dark based on `org.gnome.desktop.interface color-scheme`, and renamed the user-facing "Shell blur" wording to "Popup blur".

For the helper structure/reuse question: I agree the abstraction could probably help Coverflow Alt-Tab and maybe Window List later. I’d be more cautious with Dash/Panel because they already have a lot of compatibility-specific positioning behavior, so I’d rather not fold that into this PR. I’m also happy to move the helper files into a `src/components/shell/` subdirectory if you prefer that layout.

For the corners point, the PR currently has a popup-specific fallback for static pipelines without a corner effect, but I agree your proposed global hidden corner effect would be cleaner long term. I’m fine leaving that for a follow-up if you’d rather handle it after merging.

Also, sorry this PR got so big. The popup blur work ended up touching styling, animation timing, notification stacking, and defensive actor handling, so I kept those fixes together rather than leaving the feature half-polished.

MeguMario:
> For the theme integration, I adjusted the popup background override CSS quite a bit. The light style now actually lightens the popup surface instead of lightly darkening it. I also reworked the nested styling so the calendar/date menu keeps GNOME’s normal visual structure instead of adding filled backgrounds everywhere, while still restoring the current-day/selected-day highlight. Quick Settings also keeps checked toggle highlighting now, and normal quick toggle/menu surfaces use translucent backgrounds that fit better with blur.
> 
> For notifications, I removed the stacked-notification shadows that were getting clipped by the message list, added clearer stacked backgrounds, and fixed some light-mode contrast issues in notification headers/buttons.

It looks way better than before now! Much thanks!
There are some remaining style change that I want to see (tho it can be personal preferences for some)

- I like to see these part of the notification section to have a semi-transparent background to match the looks of the default (Adwaita) theme

image
image

- Some button still have a soild background so I would love if these can be semi-transparent as well
image

- At the default brightness level of 0.5, the background still has a darken look to the background when in light mode. At 1 it looks much brighter.
  - With brightness at 1
    image
 
  - With brightness at 0.5 (current default) in light mode
    image

Though from my testing, it seems that brightness does not work like brightness at all and instead it works more like an opacity slider, because at 0 you can't even see the blur effect at all, so ideally, we should make the background color a little bit brighter (or matching the default theme).

image
image

Proman4713:
Hey, I've got no technical skill to contribute, but I think I have some feedback 😅

- For light mode, lightening the background should honestly be the only way to go as a default, as it helps with contrast and will most likely be better with the bright type of content/wallpaper that will be under it.
- With the shell theme, I think it would generally do well for the entire extension to have an _option_ to automatically adapt app grid folders and such to the shell theme (possibly by also allowing the user to choose different brightness settings based on the shell theme? That could be nice). Because some users may want to just choose a single profile, Blur My Shell is very customisable, and I appreciate that, so maybe doing _too_ much automatic stuff with no way around it wouldn't be a wise choice

Overall, I'm excited just reading through this PR since it just looks so good. I hope it lands in the extension soon

aunetx:
Thanks for answering my points! I cannot yet answer everything sadly

In the meantime, just to tell you, in GNOME 46 the latest update crashes:
`SyntaxError: ambiguous indirect export: POPUP_ANIMATION_TIME @ file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell_surface_fade.js:2:9`
This is because `POPUP_ANIMATION_TIME` was only recently exported by GNOME shell. I think a simple fallback to `150` would fix the crash, except if you have a better idea?

And just, about the light/dark transitions: there is definitively something to do, but that concerns the whole extension. I think having a way to use different pipelines, blur settings, ... for light and dark theme would be nice (in the same way you implemented here), but that's out-of-scope of this PR. And I wonder if it's the best this to add, as it would be quite a big change and complicate things a bit...

On thing I would like to do is to be able to have dynamic custom effects, in a way similar to this extension: https://github.com/ryohsuke1231/liquid-glass. Once that's done, having a kind of "switch" in the settings for choosing a different pipeline on dark/light theme change would be easier, as it would be homogeneous for all components!

(i'm putting this here just to remember, but that's definitively not something to be added to this PR haha)

aunetx:
Also, it seems like the blur (dynamic or not) in popups does not disappear normally anymore (i think it was ok previously), like if it waited for the popup to disappear before fading out... (testing in GNOME 46 and with hardcoded `POPUP_ANIMATION_TIME=150`, that may change things).
[Video demonstration](https://github.com/user-attachments/assets/8db8ee34-70d4-44c9-aa8f-ed36b4ab952a)

Moreover, the background is weird for some popups, for example the "logout" popup is very similar and dark no matter which background option I choose:
image

Sorry for letting you do all the work, I don't have much time to code right now but if you feel overwhelmed I will definitively be able to work on this soon and fix some bugs myself!
And by the way:
> Also, sorry this PR got so big. The popup blur work ended up touching styling, animation timing, notification stacking, and defensive actor handling, so I kept those fixes together rather than leaving the feature half-polished.

Of course you don't have to feel sorry, that's very nice work and I love seeing that here :)

MeguMario:
> Moreover, the background is weird for some popups, for example the "logout" popup is very similar and dark no matter which background option I choose

I think this one depend on the shell theme you use, for example here's one without Light Style / Luminus disabled
image

And here with it enabled
image

It's only Logout / Shutdown / Reboot dialogs that have this but could be somewhere else idk

aunetx:
Yes, I guess Yaru does not play nice with it... I really hate ubuntu haha, but I guess this would still need to be taken into account as a lot of users would suffer this problem

krissedout:
Thanks for all the testing and screenshots!

@MeguMario @aunetx 
I pushed a few more fixes that should cover the main issues raised here:

- The GNOME 46 crash should be fixed now. The popup fade helper no longer depends on the newer `POPUP_ANIMATION_TIME` export.
- The popup blur fade issue should be fixed for both dynamic and static blur now. The blur no longer outlives the shell popup tint during enter/exit transitions. (this is what took me so long)
- I reworked the popup background styling so light mode uses lighter translucent surfaces, nested date menu sections/buttons are translucent, checked quick toggles keep the accent color but are less heavy, and notifications have clearer stacked backgrounds without clipped shadows.
- Logout/shutdown/reboot dialogs should now use the same real shell background/tint path instead of the old extra tint layer, so they should be less theme-dependent and less weird with Yaru/Luminus-style themes.

@Proman4713 
I agree that full theme-adaptive styling for every component would be nice, but I think that is larger than this PR. For now this keeps the explicit background options and the automatic light/dark behavior, while avoiding deeper theme color extraction or per-component pipeline switching.

krissedout:
Just as I pushed I found a bug in the resizing animation in places like QS power menu and notification center, I'm gonna try and fix it but it might take longer

krissedout:
Pushed a follow-up fix for the resize animation issue.

The blur surface should no longer turn transparent or shift awkwardly while QS / notification center popups resize. I also cleaned up the temporary workaround code afterwards, so this should be the final version from my side unless more testing finds something else.

Proman4713:
@krissedout That makes sense. I rely on Blur My Shell in all my Ubuntu setups, as well as in my amateur derivative project. So I will definitely try to learn more about GNOME's internals and contribute to extensions such as this one.

That said, it's probably going to take a long time before I'm capable enough, so I'll try to be more active through testing and/or bug reports. Thank you @aunetx, of course, for your work on this amazing extension.

MeguMario:
Thanks for your hard work! If it were to merge right now it would be super cool! I do have some thing left to report about but at this point is just minor nitpick or just edge cases come from interation with other extensions. I hope you don't mind.

- The date text in notification center shouldn't have a background behind it.
image
image

- Static blur fade out animation looks weird

https://github.com/user-attachments/assets/4c42fd4e-7dab-4e42-9be5-cc86a4d318e6

- Hovering on notification made it expand instead of highlighting it

https://github.com/user-attachments/assets/fd2be162-0dcf-4cb7-a7e7-65a11a233892

- Blur break when expanding notification stack

https://github.com/user-attachments/assets/b8ac1a94-32a4-49bf-bd67-502fd262905c


- Here's some nitpick regarding extension compatibility. 
  - [Quick Settings Tweaks](https://github.com/qwreey/quick-settings-tweaks/)'s flyout/overlay mode animation made the blur fell out of sync. Tho this can put in low prioriry due to the fact the fix for this mode hasn't been merged yet

   https://github.com/user-attachments/assets/4b973481-2f19-4eff-aac9-67c6b6309140

  - Tiling Shell's Snap Assistant somehow on top of the panel now. 

https://github.com/user-attachments/assets/3fb872de-2d04-4d31-9478-054d33aacf58

One small question that's offtopic for this PR, but do you think blur fading can be adapt to other blur surface as well (e.g Application blur)?

aunetx:
Well after trying it a bit, there are still some parts to improve:
- the static blur is a bit shifted to the left, but I don't think this is happening because of this PR but rather due to all the shifts I do in the GLSL effects / background image placement in order fix the various gnome-shell bugs (background disappearing with multiple monitors if the image is not shifted, glsl effects receiving 3px of margin somehow, ...) so I don't think there is a need to fix it here
  image
- when opening the notification menu, sometimes the blur is shifted a bit upwards (no matter if static or dynamic); and when closing and reopening again the shift disappears. Maybe some animation error?
  image
- the OSD blur in light mode makes it so that the bar is not visible, because it had a black background even in light mode (at least in Yaru, but even with Adwaita theme iirc). So maybe tweaking the color of the bar to have it black in this case?
  image
- no matter in light or dark mode, at least in Yaru, the quick settings buttons lost all color and differentiation whether or not they are activated or not:
  image
  (sorry for talking about Yaru, I am compelled to use it because of some ubuntu bugs but I guess blur-my-shell should try to maintain compatibility with both adwaita and yaru so most users with a "default" distribution theme would have something working!)

And well, there are the bugs/improvements raised by @MeguMario that would be nice to see fixed too! However, I generally see this PR as wayy good enough to be merged right now! So if you're OK, I could merge it, and fixes could come before releasing the next version when you feel like it's polished enough? In particular, I guess it's mostly cosmetic (css) fixes that have to be done, so that popup blur works out of the box for users having either adwaita or yaru, and in light or dark mode.

By the way, @krissedout and @MeguMario, if you are interested in sharing maintainership of this extension, I could add you as collaborators so that (for example) fixing later bugs in this PR, or the rounded corners etc could be easier! More generally I cannot spend a lot of time on this extension, and you provided very nice improvements to it so that could help the extension thrive without waiting for my input :)

aunetx:
By the way, about corners radius, there is a nice way to automatically get the corner radius for `St.Widget` instances, which would prevent some theming issues! Adapted to the `get_corner_radius(target, root_actor)` function in `shell.js`, this would give:

```js
let radius = (target.get_theme_node
              ? target.get_theme_node().get_border_radius(St.Corner.TOPLEFT)
              : null)
          ?? (root_actor.get_theme_node
              ? root_actor.get_theme_node().get_border_radius(St.Corner.TOPLEFT)
              : null);
```

However, this is not directly applicable there, because of the architecture of the returned object (which is not just the radius, but the settings properties etc). I guess this could actually be implemented directly in the `corners.js` effect, to try and adapt the corner radius automatically, and use the prescribed value only when it was not found or when a switch to force using a custom value is toggled on? This would really make sense for Dash-to-Dock and the panel too, as the user would not have to worry anymore about the rounded corners.

As a side note, when login out, there is a crash:
```
(gnome-shell:108286): Gjs-CRITICAL **: 14:18:37.733: Object St.Widget (0x5d2bbdd82c80), has been already disposed — impossible to access it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
== Stack trace for context 0x5d2bb948f1b0 ==
#0   7ffdaba2ad30 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell_blur_surface.js:1148 (23858b969c0 @ 42)
#1   7ffdaba2adf0 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell_blur_surface.js:1142 (23858b96920 @ 632)
#2   7ffdaba2aec0 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:386 (23858b8f0b0 @ 124)
#3   7ffdaba2af80 b   file:///home/ahamy/.local/share/gnome-shell/extensions/blur-my-shell@aunetx/components/shell.js:183 (23858b8aab0 @ 18)
#4   5d2bbdd96898 i   resource:///org/gnome/shell/ui/layout.js:247 (3b5c0b5df560 @ 328)
```
It is not very important as I did not see it before shutting down, but may as well be fixed!

aunetx:
Another note: some tool was just released to inspect and profile gnome extensions: https://github.com/todevelopers/gse-profiler.

I did not test it too much but it looks promising, and after inspecting a bit the impact of this PR, it looks like it is quite heavy in some calls (that happen when switching workspaces for example, but not opening the popups), for example `find_target_children` and `try_blur`:
image

More profiling may be needed, but if you have any idea to solve some performances problems at least that is quite a helpful tool! And I will spend some time profiling the entire extension later anyway, that may help find memory leaks too :)

krissedout:
Alright, this took a bit, but I’ve refactored the popup blur implementation to be more organized and hopefully more performant.

The popup blur code now lives under `src/components/popup`, and I renamed the previous files/settings around "shell blur" to "popup blur" so the naming matches what this code actually handles. This should also fix the anchoring-related bugs I introduced and the issues you mentioned.

One note: since the preferences were renamed too, this will likely reset/break existing popup blur settings. I think that is worth doing now rather than keeping the old naming around after this lands.

I’d appreciate another test pass from you as well. I also would not merge this yet until we’re reasonably sure the popup blur behavior is stable, especially around performance and transitions.

---

@MeguMario

> One small question that's offtopic for this PR, but do you think blur fading can be adapt to other blur surface as well (e.g Application blur)?

Yes, blur fading can probably be adapted to application blur, but I think that should be a separate follow-up. Popup fade is actor/geometry based, while application blur is window based, so it likely needs its own per-window fade path in `applications.js` rather than being a direct drop-in reuse.

Also, somewhat related: I opened an upstream [Mutter MR](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/5071) implementing support for the new Wayland ext-background-effect-v1 protocol. If that eventually lands, it could give Wayland clients a native compositor-side blur path, which may simplify some dynamic/application blur cases in the future.

That is obviously separate from this PR and not something Blur my Shell can rely on yet, but it may be useful context for the longer-term direction.

---

@aunetx

> By the way, about corners radius, there is a nice way to automatically get the corner radius for St.Widget instances, which would prevent some theming issues! 

Good point, we can probably use the theme node radius as the default source and keep the configured values as fallback/override. I don’t think it should fully replace the settings because some actors don’t expose useful radius values, and users may still want to force a custom value, but using get_border_radius() first should reduce theme-specific corner mismatches.

> By the way, @krissedout and @MeguMario, if you are interested in sharing maintainership of this extension, I could add you as collaborators so that (for example) fixing later bugs in this PR, or the rounded corners etc could be easier! More generally I cannot spend a lot of time on this extension, and you provided very nice improvements to it so that could help the extension thrive without waiting for my input :)

Thank you! I’d love to help maintain it.

I’m still getting familiar with the whole extension, but I’m happy to help with the areas around this PR, popup/shell blur, styling polish, rounded corners, theme compatibility, and later bug fixes/regressions related to those changes.

MeguMario:
@krissedout 
Nice! Thanks for the update, it's always excited to see the inbox having new update from this PR.

> Yes, blur fading can probably be adapted to application blur, but I think that should be a separate follow-up. Popup fade is actor/geometry based, while application blur is window based, so it likely needs its own per-window fade path in `applications.js` rather than being a direct drop-in reuse.

Neat, I was banging my head trying to find a way to reuse it but seems like there's more to this than I've anticipated. Thanks for the answer though!

> Also, somewhat related: I opened an upstream [Mutter MR](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/5071) implementing support for the new Wayland ext-background-effect-v1 protocol. If that eventually lands, it could give Wayland clients a native compositor-side blur path, which may simplify some dynamic/application blur cases in the future.
> 
> That is obviously separate from this PR and not something Blur my Shell can rely on yet, but it may be useful context for the longer-term direction.

Well this could take years if we lucky but I hope to see it get merged into mutter as well. With this dynamic blur can be much simpler as we don't need to have an external dependency anymore.

There are still a few more from my testing (most of them are later addition to the comment I made earlier so, I can just... cut and paste them here and we are good). 

- Volume control also shouldn't have a background behind it (slider and buttons)

Screenshot From 2026-05-22 13-19-30
Screenshot From 2026-05-22 13-19-36

- Some submenus are having their background opaque (e.g IME settings)
image

- Same for some surface though this could just be something that extensions have to account for (below is ArcMenu)
image

- While the blur tracking is fixed now, sometime the background is a little bit shorter after the animation finished (can be seen with Quick Settings Tweaks's flyout mode)

https://github.com/user-attachments/assets/c49dcd3a-4d52-4e15-a4ac-a76442854a12

There are also issue where static blur is still fade to black and tiling shell's quick assistant leave the blur surface on top instead of below the panel so I would love to see one of them (or all of them) get fixed as well

@aunetx 

> By the way, @krissedout and @MeguMario, if you are interested in sharing maintainership of this extension, I could add you as collaborators so that (for example) fixing later bugs in this PR, or the rounded corners etc could be easier! More generally I cannot spend a lot of time on this extension, and you provided very nice improvements to it so that could help the extension thrive without waiting for my input :)

I'm love to help out as well, though it may take some time for me to familar myself with the extension though.

krissedout:
@MeguMario 
I pushed another round of fixes for the issues you listed here:
- volume sliders/buttons should no longer draw the extra background
- IME/submenu backgrounds should be handled better now
- quick settings toggle/menu-button states should now match GNOME’s behavior more closely
- the blur surface tracking/animation sizing should be more stable after the transition finishes

I also split `stylesheet.css` into smaller files under `src/styles`, since it was getting too large to maintain comfortably.

For ArcMenu, the issue is that it uses its own style classes for some surfaces, so we would need to hardcode ArcMenu-specific overrides in Blur my Shell. That is something we could do, but I don’t think it would be a good idea right now. I’d rather avoid extension-specific styling unless we decide that we should maintain compatibility overrides for popular extensions in general.

I can't repro the static blur fade-to-black issue, but that could've gotten fixed with these changes too.

MeguMario:
@krissedout 
Thanks for the swift reply, unfortunately this round of fixes also broke background override in other components (Overview and Panel) with light mode broke more than dark mode. So I assume the stylesheets need to be adjust again.

image

Otherwise, here are more report from what I've found:

- The two bottom corners of submenu background are unrounded

image
image

- While static blur fade effect did not fade to black anymore, it also last longer than expected and can be seen easily on apps

https://github.com/user-attachments/assets/b6195f2c-b4b1-4d5b-88e8-bbe635155f89

- The Clear button is opaque

image

- Inconsistent QS buttons background
image
image

Sorry for piling all of this on you, I hope you don't mind. 
Thanks for your hard work!

Off-topic but I found that this PR can be used to blur the dock that have the classes of `dash-background` and `plank-like-dock` (That mean Dash-to-dock, Dhruva, Dock!ng, Bottom Dash Panel,...). Though it probably should be something seperate from this PR 

image
image
image
image

krissedout:
@MeguMario 
Thanks again for the detailed testing. I pushed another fix pass for the latest reports:

- fixed panel/overview background overrides being broken by the popup stylesheet changes
- restored the submenu bottom corner radius
- fixed the static blur fade lingering too long after the popup is hidden (I don't know if this is fixed as I don't have this issue, but it should be)
- made the notification Clear button follow the popup background style
- kept the QS button background fixes from the previous pass

MeguMario:
@krissedout Thanks for the hard work. It seems like we are inching closer to perfection here (I think).

While static blur fade did not lingering after the popup is hidden, it also fade to black again. After some investigation, I found out that any texture effect (the exception are Pixelize and Luminosity though Luminosity looks kinda funny when you hide the popup) that you can add in static blur pipeline can cause that to happen, remove those should make it not fade to black anymore. So I guess we have to make sure that the texture effect is also fade out as well

https://github.com/user-attachments/assets/1785399a-7db7-4fb1-b385-712409a7a733

Also the IME submenu still have those corners squared but only visible on bright background
image
image

Blur can appear for a few second on second display when an element is suppose to go out of the screen they displayed on and hide

https://github.com/user-attachments/assets/beb7dcab-0505-4a79-a10e-1601d5403c4f

Or you like me and you are playing around, then it is there all the time if it is a dock (or other object that do not hide themselves and simply just out of view of the current monitor)

https://github.com/user-attachments/assets/2e060e1d-4bfe-468b-aab6-87a1848d87b5

I think we should implement a check where the effect will fade out if an object ever go out of screen.

The background app indicator inside QS also shouldn't have a background, but it looks rather good with it so I'll leave it to you
image
image



Outside of those beside from QS button that have submenu having more opaque background compare to one that does not have it. I think I don't have much more to report, so if I have something to say I will make another post. Otherwise on my part it's good to go!

MeguMario:
Late addition, but I think the notification background could be a little bit brighter, tho it could intefere with the way notification stack is styled so hard to find a solution to this one.
image
image

krissedout:
@MeguMario 
I made another pass on the remaining issues:

- Static blur fade: texture effects now fade with the static surface opacity instead of staying fully active, so Noise/Color/etc should no longer make the popup fade to black.
- Off-screen surfaces: popup blur is now clipped to the monitor where it appears, and if the actor moves fully outside that monitor the blur hides instead of keeping the cached surface visible.
- QS buttons with submenus: fixed the extra opacity by no longer painting the split-toggle parent background, only the actual button segments.

For the IME submenu corners, I tried to cover this with the submenu radius rules, but if it still reproduces after this I’ll handle that specific row/header case next.

For the background app indicator and notification brightness, I left them alone for now. The app indicator looks acceptable to me, and notification stacks are pretty sensitive, brightening the notification cards globally can make stacked notifications look too opaque, especially in dark tint mode.

MeguMario:
@krissedout

> * QS buttons with submenus: fixed the extra opacity by no longer painting the split-toggle parent background, only the actual button segments.
> 
> For the IME submenu corners, I tried to cover this with the submenu radius rules, but if it still reproduces after this I’ll handle that specific row/header case next.

image

Unfortunately it is still reproduced

Outside of that I think there's not much more to report, aside from this

image

Can the scrollbar be semi-transparent I wonder.

With that I think it's everything I could find (I hope). There's also compatibility with Yaru theme which @aunetx or anyone with Ubuntu can help out with. Thanks for the hard work!

krissedout:
@MeguMario 
Fixed these too

For the IME submenu, the remaining square corners were coming from the inner submenu section/row styling rather than only the outer popup-sub-menu, so I added radius coverage for those nested sections too.

For the notification center scrollbar, I made the scrollbar handle semi-transparent generically under the popup background override, so it should also apply to other scrollable transparent/tinted popup surfaces and not only the date menu.

MeguMario:
@krissedout 

> For the IME submenu, the remaining square corners were coming from the inner submenu section/row styling rather than only the outer popup-sub-menu, so I added radius coverage for those nested sections too.

I think it doesn't work. Either the stylesheet didn't apply or there's another layer behind it.

image

Another unrelated point, so somehow one of the fix pass also fix the background tracking for object that have its size change dynamically (The last time it stuck in one place) so kudos for that.

https://github.com/user-attachments/assets/96426a75-eecb-4b2d-b981-586ea3b0c188

Sorry for the color, I was recording on an HDR enabled monitor 

Thank you so much for your hard work by the way!

krissedout:
@MeguMario try again please!

MeguMario:
@krissedout ok so the entire background is gone now 

image

krissedout:
@MeguMario that was an overcorrection from me, can you try now?

MeguMario:
@krissedout okay it is fixed now, though it is a little hard to see
image

And depending on the background it's basically invisible
image

Can you make it a little bit brighter, it should be OK after this.

krissedout:
@MeguMario done

MeguMario:
@krissedout ok, all fixed now! thank you so much!

MeguMario:
I hope this is the final one but can you make the date text when hover have a semi-transparent background. Somehow I missed this.

image

aunetx:
Hi! Thanks a lot for all the improvements, the re-structuration of the code and the style fixes are very nice! And I checked again the perfs profile, and there is no issue anymore about it!!

If I understood correctly, everything seems to work well on GNOME 50 now (except for the date text hover style issue haha)?

However I have some issues on GNOME 46, that appear no matter if I have Yaru or Adwaita shell theme (I was able to install it to test it):
- the "automatic" theme setting is actually reading the "entire" GNOME theme setting, and not the specific GNOME shell theme: by default, in Adwaita even with a light theme the GNOME shell theme is dark, and the light theme is activated by this extension for example: https://extensions.gnome.org/extension/6198/light-style/. In Yaru, setting a light GNOME theme also switch the GNOME shell theme to light. So I guess having a way to have the "Automatic" setting somehow figure out the real GNOME shell theme would be perfect?
  - for Adwaita, reading `Main.sessionMode.colorScheme` and applying the light style only if it is set to `prefer-light` should do the trick
  - for Yaru, they do not respect this setting (it is always `prefer-dark`), and simply use the "global" GNOME theme to set the light or dark Yaru shell theme. So in this case, keeping it as-it would be OK!
- the OSD in light mode is dark (with dark content so it is not readable), and even in dark mode it looks like it is not blurred either and pitch black
  image
- the all-screen popups have some issues:
  - logout screen popup is not fully blurred 
    image
  - `alt+f2` popup is not blurred at all
    image
- no matter in light or dark mode, and with adwaita or yaru, the buttons that were meant to use the accent color (that does not exist in GNOME 46) do not use it, and do not fallback to the original blue accent color. This also means that buttons that were meant to use the accent color and are hovered have a weird text color, however I guess that will be fixed once the right background is set
  image

But except for this, this is perfect! So I guess once this can be fixed, this could be merged? And after that, I'll implement automatic corner detection for the whole extension, and maybe a bit of work on dash-to-dock and panel blur positioning would be needed at some point (because there are a lot of issues about it with multi-monitors...); before publishing the next version!

And thanks for agreeing the shared maintainership, I'll add you right now, and add some branch protection for the `master` branch to require at least a review before merging (I guess with multiple maintainers this is the better way to proceed?)!

krissedout:
Thanks, I handled those too!
@MeguMario
- Date menu today/header hover now uses a translucent background in dark mode too.

@aunetx 
- Automatic popup style now uses `Main.sessionMode.colorScheme` when GNOME exposes the real shell color scheme, with a Yaru fallback through the interface theme because Yaru reports the session mode as dark.
- GNOME 46 now has explicit accent color fallbacks for checked quick settings buttons/calendar today before using the newer `-st-accent-*` colors.
- `modal-dialog` and `run-dialog` now use full-geometry popup blur/ background handling, so this should improve the logout and Alt+F2 cases.

I can’t fully verify the GNOME 46-only cases locally, so a retest there would be useful.

MeguMario:
@krissedout 

> * Automatic popup style now uses `Main.sessionMode.colorScheme` when GNOME exposes the real shell color scheme, with a Yaru fallback through the interface theme because Yaru reports the session mode as dark.

Unfortunately this broke automatic theme detection on GNOME 50. Automatic detection will stuck in dark mode even in light mode, and with Light style enabled enabled, vice versa. 

image

@aunetx 

> * for Adwaita, reading `Main.sessionMode.colorScheme` and applying the light style only if it is set to `prefer-light` should do the trick

On stock GNOME, `prefer-light` is pretty hidden so `default` is assumed to be light mode. So it basically result to `default` for light mode and `prefer-dark` for dark mode if you change it via Settings, so I assume it also need to check for `default` in addtion to `prefer-light` as well (or in reverse, only apply dark mode if `prefer-dark` is set)

krissedout:
@MeguMario try again please, I'm not testing locally as I have other work so I can't log out and back in right now, I hope this works tho!

MeguMario:
@krissedout 
It seems like it does not work as having Light style enabled keep the shell in light mode even in dark mode (It's forcing prefer-light even in dark mode ) [link](https://gitlab.gnome.org/GNOME/gnome-shell-extensions/-/blob/main/extensions/light-style/extension.js?ref_type=heads). While checking the extension, ~~while Light style does restore the previous color scheme, our extension got that value before Light style get to restore the color scheme and stuck in prefer-light, so a little delay to wait for light style to restore the previous value should be it~~ it seems to force `prefer-light` regardless if dconf is set to `prefer-dark`

edit: correction

aunetx:
As indicated by @MeguMario, there seems to still be some problems about theme switching: when using Yaru and dark mode, the popup theme is set to dark (even though it should stay light because yaru); and with Adwaita the theme is not updated when light/dark mode is changed (it only follows the `Main.sessionMode.colorScheme` setting, but without listening to its change so it updates when the theme mode itself is changed).

However the rest has been fixed,
- except for the OSD popup that still has a pitch black background (looks like the blur does not work, it was working some iterations ago although with wrong styling, now at least it has the right styling but no blur!)
- and a tiny detail: in light mode, the arrow of buttons that are both activated and focused becomes white even though it should stay dark
  image

krissedout:
@MeguMario @aunetx I've pushed some changes, it should hopefully fix some of the bugs (hopefully all), though it seems when I fix one 10 new ones take it's place haha

krissedout:
After this PR is merged, I think we should look at centralizing the shared blur logic. Existing configs can stay compatible, but instead of each surface maintaining its own blur implementation, we could have one reusable blur layer and let panel/dash/popups/apps decide where it applies. Popup blur already moves in that direction, so I’d like feedback on whether this sounds like the right follow-up.

MeguMario:
@krissedout 

unfortunately the autodetection doesn't work, tho it seems like the light style force `prefer-light` regardless of user settings so we have to make special case for that

image

I wonder if the Not responding dialog can be blurred as well
image

> After this PR is merged, I think we should look at centralizing the shared blur logic. Existing configs can stay compatible, but instead of each surface maintaining its own blur implementation, we could have one reusable blur layer and let panel/dash/popups/apps decide where it applies. Popup blur already moves in that direction, so I’d like feedback on whether this sounds like the right follow-up.

As for this, personally, I think the foundation made by this PR already handle Dock blur quite well, it seems to not exhibit some bugs that the current implementation has (specifically https://github.com/aunetx/blur-my-shell/issues/844) while having space for a more generic support for more dock extension (since it looks for style classes instead of child name), in addtion, it may make maintaining a bit simpler as well.

aunetx:
@krissedout thanks for the changes, I cannot test now but indeed there are a lot of additional fixes for each bug fixed! That's why I guess merging quickly would help have a "stable" interface, and if that's only cosmetic changes that would be OK to do in another PR with further and more generalized testing.

And about a more general implementation of blurring for other actors, I'm fully positive with you, having something that can be reused for every kind of surface would be perfect... that was a bit my idea with the pipelines (it abstracts a bit of the complexity away from the modules in `src/components`), but going further with your abstraction would help making the extension more robust.

Something I would love to test too is to have something similar to https://github.com/ryohsuke1231/liquid-glass for dynamic blur: the method of this extension would allow to have dynamic effects without the need to use external libraries, so that would be a nice thing to group the static and dynamic blur in only one abstraction. But that will probably not be easy, and that especially may be a bit slower than the current dynamic effect...

krissedout:
@MeguMario 
I added a special case for GNOME's Light Style behavior, since it forces `Main.sessionMode.colorScheme` to `prefer-light`, which makes normal autodetection unreliable there.

I also added tracking for the window-attached `close-dialog`, so the “Not Responding” dialog should be picked up by popup blur now. That one would still be good to retest because it is attached under the app window actor rather than the usual shell popup groups.

@aunetx 
Yes, I agree. I think we should try to merge this PR soon once the core behavior is confirmed to work reliably, since the PR is getting quite large now and we are already close to 60 comments. At this point it would probably be cleaner to use this PR as the stable base, then handle the remaining style/cosmetic issues and further refactors in follow-up PRs. That should make review and testing much easier, instead of continuing to expand this one indefinitely.

For the more general blur abstraction, I agree as well, but I think the next step should probably be a dedicated follow-up PR where we abstract the static and dynamic blur paths more cleanly and make it reusable for other actors/surfaces too. Regarding something similar to `liquid-glass`, that sounds interesting. I agree it would probably be more complex, however I will explore that too.

MeguMario:
@krissedout 

It seems like you switched to `Main.getStyleVariant?.()` for detection, Light style detection is no longer necessary because they always return the correct state unlike `Main.sessionMode?.colorScheme` (`dark` and `light` with Light style enabled, and only `dark` with it disabled since it is the default state of gnome's shell style)

Regarding the "Not responding" dialog, it seems to work now
image

Also, there's a small merge conflict so you might want to resolve it

aunetx:
Well, it looks like `Main.getStyleVariant()` indeed works all the time, sorry for talking about `Main.sessionMode.colorScheme` which was not the right setting to look at! So only using that setting (and a Yaru exception, because even the true dark mode of Yaru is light...) would work perfectly if I understood correctly.

About the OSD not being blurred, I realized that my problem was coming from Just Perfection, whose theme actually overrides the background color and makes it pitch black. I tried to fix it but was not able to, I don't know why Just Perfection's theming takes precedence over Blur my Shell's even with more precise css selectors, it does not seem obvious to me...

Sorry about the merge conflict by the way, I merged another PR before this one but did not think it would conflict. I have some changes to push too (regarding something else completely), so if you feel OK, once the theme issue selector is solved we can merge that and iterate from there!

krissedout:
@aunetx @MeguMario Thanks for the review and about merge conflict, I merged upstream locally and didn’t hit any unresolved conflicts on this branch, so there shouldn't be any now.

Style variant detection: `get_background_style()` now uses `Main.getStyleVariant?.()` as the primary signal (no longer uses `sessionMode.colorScheme`), with the interface `color-scheme` setting as fallback only, also removed the Light Style extension check path, since it’s no longer needed.

OSD theme override issue: The OSD target now gets an explicit background override at runtime (inline style), so `osd-window` backgrounds are forced to stay transparent and don’t get forced black by theme-level selectors (notably Just Perfection) that were taking precedence. This avoids fighting global stylesheet order and only affects OSD targets we blur. I haven't tested this, but if it still doesn't work that's an extension compat issue and I think it doesn't have to be addressed in this PR yet.

MeguMario:
@krissedout 

While the OSD is no longer a solid black like before with Just Perfection, it's also do not follow BMS background override either (light mode still have the OSD with dark background), this however still occur with or without Just Perfection

image

krissedout:
@MeguMario good catch, try again please!

MeguMario:
@krissedout I was editing the post then I see the update (lol)

- The color effect for static blur seems to only effect popup blur for some reason
image

MeguMario:
- The OSD was fixed

image
 
- It seems like Just Perfection also override the window switcher style as well

https://github.com/user-attachments/assets/f423eb7d-4e80-4fc3-a979-3e4617c9ea87

- Also I think we forgot to test the Transparent background option

image

aunetx:
Thanks for the fixes guys and sorry I cannot test them right now; but it seems like the only changes left are CSS right?

If so we can merge it by tomorrow noon (french timezone hehe), so it gives us more space to iterate from there. If the css is fully fixed by then it's perfect, else it will entirely be doable by a follow-up PR! And this would allow us to ask for some testing by other users, I am especially interested in knowing whether or not it fully works on GNOME 45 for example

aunetx:
Well, I could test it, and I confirm that everything with adwaita theme works perfectly!

However, the Yaru theme breaks that, as its dark variant is indeed not dark. I tried to find a way to know whether the current gnome shell theme is a light-style Yaru theme, and the only way I did find is:

```js
let theme = St.ThemeContext.get_for_stage(global.stage).get_theme();
let should_stay_light = false;
if (theme.application_stylesheet) {
  let uri = theme.application_stylesheet.get_uri();
  should_stay_light = uri.includes("Yaru") && !uri.includes("dark");
}
```

I don't know exactly where to put it in particular (and it might be worth it to listen for gnome theme change somehow), but at least respecting the `should_stay_light` var would allow the theme chooser to work seamlessly for ubuntu users!

krissedout:
@aunetx I merged the Yaru auto-detect fallback into popup style resolution inside `get_background_style()`.

@MeguMario
For now, I don’t think this PR can reliably support conflict cases with Just Perfection’s style overrides or GNOME Light Style. The core issue is that Blur My Shell can provide the background blur/tint layer and keep its own popup styling consistent, but it can’t reliably interoperate with every other extension that also modifies Shell CSS.

Those extensions/themes can force their own background or color rules with higher CSS precedence, so once multiple extensions are trying to style the same Shell components, the result becomes order-dependent and hard to reason about. Properly handling all of those combinations would expand the scope quite a bit and likely require extension-specific compatibility handling.

This is also why I think targeting base GNOME/Yaru style compatibility is the right scope for this PR. BMS already overrides a fair amount of Shell styling to make blurred components work correctly, including light-mode handling, so mixing it with another extension that tries to restyle those same components can create conflicts outside the core popup blur/tint logic.

If this is acceptable, please test with those extensions disabled. For Just Perfection, disabling only the style override should be enough. If needed, we can handle compatibility for those cases in a follow-up PR.

> Also the color effect for static blur seems to only effect popup blur for some reason

I think that should be tracked and investigated separately. I didn’t touch that rendering path in this PR, so I’d rather not bundle that into this change. If needed, I can take a look at it in a follow-up PR once this lands.

I also removed the inline OSD/switcher background overrides from JS, so it’s no longer fighting theme CSS at runtime. That was mostly a hack to force style precedence, and if we still need something like that, it should probably be handled separately in another PR.

aunetx:
@krissedout thank you for the quick fix! There are two problems:
- the blur disappeared on every popup; the styling is still set and I don't see any logs... To be honest I don't really know what could have broken it :/
  image
- the Yaru theme detection actually only applies on `gnome-session` and not on the default ubuntu session (that uses Yaru as the default theme, and not the application stylesheet). So a complete detection function would be:
  ```js
  is_yaru_light_theme() {
      try {
          const theme = St.ThemeContext.get_for_stage(global.stage).get_theme?.();
          let uri = theme?.default_stylesheet?.get_uri?.();
          if (theme.application_stylesheet)
              uri = theme?.application_stylesheet?.get_uri?.();
          return typeof uri === 'string' && uri.includes('Yaru') && !uri.includes('dark');
      } catch (e) {
          return false;
      }
  }
  ```
  (I kept the `?` and the `try catch` you introduced, but to be honest I'm not sure they are all needed here? this seems a bit redundant)

krissedout:
@aunetx Thanks a lot for the testing + for flagging this.

You’re right that in Ubuntu sessions the theme can come from either application_stylesheet or default_stylesheet, so your detection idea is correct. I also found a small bug in my previous helper (I only checked one stylesheet), so I’ll switch to a safe fallback version using both and a case-insensitive check.

I think the ?. + try/catch is fine to keep here for safety.

For the “blur disappeared on all popups” issue, I accidentally removed the has_any_style_class() which broke it, I undid that it should work now.

aunetx:
Thanks for the fix, blur now works fine. However the logic for yaru detection is a bit off: you flag on `uris.some`, but it should be flagged in order (because the default stylesheet may be yaru, but the one actually used is the application stylesheet if it exists).

Moreover, there is a catch: if `Yaru-dark` is used, then we should also not switch to the light theme. So a correct `get_background_style` function would be (I removed the `is_yaru_light_theme` because it complicated the logic handling):

```js
get_background_style() {
    const style = this.settings.popup.STYLE_POPUP;
    if (style >= 0 && style < POPUP_BACKGROUND_STYLES.length)
        return style;

    const theme = St.ThemeContext.get_for_stage(global.stage).get_theme?.();
    let uri = theme?.default_stylesheet?.get_uri?.();
    if (theme.application_stylesheet)
        uri = theme?.application_stylesheet?.get_uri?.();
    const lower_uri = typeof uri === 'string' ? uri.toLowerCase() : '';
    if (lower_uri.includes('yaru'))
        return lower_uri.includes('dark') ? 2 : 1

    const shell_style = Main.getStyleVariant?.();
    if (shell_style === 'light')
        return 1;
    if (shell_style === 'dark')
        return 2;

    return this.interface_settings.get_string('color-scheme') === 'prefer-dark' ? 2 : 1;
}
```

krissedout:
@aunetx done!

aunetx:
Yayy, works perfect! Now the OSD problem with just-perfection reappeared though, but if you prefer to leave it like that i'll understand (even though at some point it would have to be fixed, Just perfection has nearly 2m downloads so we can't afford to break the sound OSD for its users haha)

But i'll be good for merging this, if @krissedout and @MeguMario you are ok :)

krissedout:
Works for me. I’d prefer to leave the Just Perfection style issue for a follow-up PR so this one can merge without expanding the scope again, since from what I’ve tested it affects more than just the OSD and needs a deeper look through JP’s style classes. For now, disabling the style override in Just Perfection should avoid the conflict.

After this is merged, I (or someone else) can take a separate look at the JP/style-override compatibility path, since that probably needs more targeted handling than the base popup styling fixes here. Also if you're going to merge it, you should likely squash and merge as the history is kind of a mess here :)

MeguMario:
@krissedout I understand as this PR has already gone long enough (lol), though I'd liked for the static blur problem (it being texture effect not being apply correctly outside of popup blur) to be at least fixed before that though, since that one is independent from any styling and extension related problem and possibly come from the fade to black issue with texture effect we've talked about before.

Though other than that, no problem merging it right now from me.

krissedout:
@MeguMario So sorry about that, I pushed a fix that should work (keep in mind I haven't tested it so please tell me if it still doesn't work)

MeguMario:
@krissedout 

Most of them still do not work, pixelize effect result in this however

image

aunetx:
I'll wait for this before merging finally!

There is still an important question anyway: should we enable popup blur by default for users? And if so, with static or dynamic blur?
Because dynamic blur works very very well for me, but that is because I have the rounded corners library. Enabling popup blur with square corners would not be doable (we should not expect users to do something to have a nice setup after a simple extension install); enabling it with static blur is a bit sad; and disabling it by default is even worse.

So maybe enabling it by default if and only if the user has rounded corners library installed? Or else, we can wait before releasing the next version for a true dynamic blur method that would not require an external library (similar to https://github.com/ryohsuke1231/liquid-glass).

krissedout:
@aunetx @MeguMario Update: I figured out a fix.

Sorry this took longer than expected. I had to spend some time understanding/debugging the shader side properly, but it now works on my end and no longer throws the `GL error (1282): Invalid operation` error.

While fixing it, I also ended up doing the refactor I mentioned earlier. The blur logic is now centralized in a shared `BlurSurface`, and the popup blur, shell surfaces, and related blur code have been moved over to use that shared implementation.

This also fixed several bugs along the way. For example, the popup blur already had a corner shader effect, so I moved that into the shared `BlurSurface` instead of keeping it popup-specific. (which means we shouldn't need the external dependency anymore 🎉)

I also changed the preferences UI as part of this work, mostly because the old structure did not map very cleanly to the new shared blur implementation.
image

I think there are three possible ways forward:

1. I can isolate only the shader fix, keep this PR focused on popup blur, and then open a separate PR for the shared `BlurSurface` refactor and preferences changes.
2. I can push the current refactored work here, and we can merge the popup blur fix together with the shared blur implementation.
3. This PR can be merged as-is, and I can open the refactor as a follow-up PR, since that version fixes the remaining related bugs more broadly.

Option 3 would be the easiest on my side because the fix and refactor are fairly intertwined now, but I am fine with whichever approach you prefer.

MeguMario:
@krissedout
I'm with option 3 as well, seeing that this refactor goes beyond the scope of this PR (and being probably the biggest change this extension has ever seen), so it's probably best that these are seperated

Excited to see the changes!

aunetx:
Hi, sorry for the delay, option 3 seems the best on my side too! So everybody is OK to merge it right now?

By the way, the `BlurSurface` sounds very exciting! I'm actually puzzled by how you managed to fix the corner thing, you mean users would not need to install the external rounded-blur library? If so, that's super nice!!

And about the preferences, I think we will need to think about it a bit, because centralizing it the way you did is indeed the simpler way technically speaking, however wouldn't it make it more difficult for users to adjust the behaviour of some parts of the extension? I mean I'm telling that but that's not for this PR anyway, we'll see that later!

aunetx (approved):

krissedout:
Alright, sounds good! I’ll merge this PR as-is then and open the refactor as a separate follow-up PR soon (I have a few last polishes yet to do), there are indeed quite a lot of changes in that now, but I think you might be pleasantly surprised by some of them :)

PR #892, opened 2026-05-26
fix pipeline.js: prevent wallpaper swap flash on blurred surfaces

EvianMcKeown:
### Description of Issue
When the wallpaper changes (slideshow background change, might be other cases) static blur surfaces briefly show a solid background colour before the new wallpaper appears. The flash is consistently reproducible in the overview.

### Cause
[`BackgroundManager`](https://github.com/GNOME/gnome-shell/blob/7bc1a00733cab6d1da619a43fb4d4c3816ff7799/js/ui/background.js#L684) is constructed with `controlPosition: false`, disabling the sibling re-ordering that `BackgroundManager` normally performs in [`_createBackgroundActor()`](https://github.com/GNOME/gnome-shell/blob/7bc1a00733cab6d1da619a43fb4d4c3816ff7799/js/ui/background.js#L801-L805): when a new `MetaBackgroundActor` is created during a wallpaper swap, it lands on top of the existing actor rather than below it.

While the new actor's image is loading, `MetaBackgroundContent` displays `DEFAULT_BACKGROUND_COLOR`.

### Fix
Reinstate the sibling re-ordering in a `child-added` handler. New `MetaBackgroundActors` are moved down in the container, so their 'loading' frames are hidden behind the existing actor.
GNOME's 1s opacity fade of the old actor then produces a blurred crossfade with no visible artefacts.

```js
this.actor.connect(
    'child-added', (_container, child) => {
        if (child instanceof Meta.BackgroundActor)
            _container.set_child_below_sibling(child, null);
    }
);
```

### Testing
Tested on GNOME Shell 50, Wayland, with a slideshow cycling every 30s:
- Wallpaper change while on the desktop: smooth blurred crossfade → no flash
- Wallpaper change while in the overview: previously the most reliable reproduction → no flash

### Notes
A reliably reproducible flash in the overview's unblurred workspace background (not the Blur-My-Shell blur layer) is present with all extensions disabled. That is a separate upstream issue in `gnome-shell` itself and is out of scope here.
I'm currently working on a gnome-shell patch for that separately.

### DIFF
--- a/src/conveniences/pipeline.js
+++ b/src/conveniences/pipeline.js
@@ -69,7 +69,7 @@ export const Pipeline = class Pipeline {
         // sibling re-ordering.
         // Without it, new actors render on top while loading, causing a solid color flash through
         // on the surface.
-        this.actor.connect(
+        this._child_added_id = this.actor.connect(
             'child-added', (_container, child) => {
                 if (child instanceof Meta.BackgroundActor)
                     _container.set_child_below_sibling(child, null);
@@ -148,7 +148,10 @@ export const Pipeline = class Pipeline {
         this.remove_all_effects();
         if (this.actor && this.actor_destroy_id)
             this.actor.disconnect(this.actor_destroy_id);
+        if (this.actor && this._child_added_id)
+            this.actor.disconnect(this._child_added_id);
         this.actor_destroy_id = null;
+        this._child_added_id = null;
         this.actor = null;
     }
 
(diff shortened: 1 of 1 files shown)


aunetx:
Thanks a lot for the PR! It looks like you never disconnect the signal from the actor, maybe it should be done at some point? Except if there is a good reason for not disconnecting (in which case you could point it out in a comment so that it passes extensions.gnome.org reviewing); I don't remember if the actors created with `create_background_with_effects` can outlive the scope of the pipeline...

EvianMcKeown:
Pushed an explicit disconnection in remove_pipeline_from_actor(), following the same pattern as actor_destroy_id. Does that seem like the right place to you?

aunetx (approved):
That's perfect thank you! I renamed two symbols to match the rest of the code and I'll merge that :)

PR #893, opened 2026-05-30
Add Liquid Glass-style refraction effect

treebarkbr:
## Summary

Adds a new Blur my Shell pipeline effect based on the iOS 26 Liquid Glass look. The shader is directly adapted from winaviation's Liquid (Gl)ass tweak, specifically `Shared/LGMetalShaderSource.m`, and ported from Metal shader code to GLSL for Blur my Shell

Source: https://github.com/winaviation-tweaks/liquidass

## How it works

The effect plugs into the existing Blur my Shell effect pipeline and samples the current pipeline texture behind the actor. From there, the shader uses a rounded-rect distance field to find the glass edge, corner radius, and outward normal. That edge data drives a Snell-style refraction profile so pixels near the border are displaced like they are passing through a beveled piece of glass

The shader also applies a small internal blur pass, optional RGB fringing, tint, inner shadow, and a highlight pass based on backdrop luminance so brighter wallpaper areas can catch the rim a bit more naturally

## Why draft

Opening this as a draft because the effect is working, but there are still a couple of visual issues I want to sort out before marking it ready for review

I am still looking into the known issues below. Suggestions or pointers are appreciated, especially from anyone more familiar with the effect pipeline or Dash to Dock

## Known issues

- 45 degree seams can appear on rounded corners when the refraction effect is applied. This is much less noticeable on smaller elements
- ~~Dash to Dock has a weird shifting effect when item magnification is enabled~~ Was caused by my fork of Dash to Dock.

## Notes

The latest pushed change only formats the refraction shader and effect wrapper for readability. The main behavior is from the earlier refraction effect work

## Testing

- Tested manually on GNOME 50, Wayland
- Checked whitespace/errors in the diff with `git diff --check`

### DIFF
--- a/src/effects/effects.js
+++ b/src/effects/effects.js
@@ -401,7 +401,7 @@ export function get_supported_effects(_ = () => "") {
                     description: _("Blurs the sampled backdrop before the liquid-glass shader is applied."),
                     type: "float",
                     min: 0.,
-                    max: 80.,
+                    max: 48.,
                     increment: 1.,
                     big_increment: 10.,
                     digits: 0
@@ -415,6 +415,16 @@ export function get_supported_effects(_ = () => "") {
                     increment: 1.,
                     big_increment: 10.,
                     digits: 0
+                },
+                rim_width: {
+                    name: _("Rim spread"),
+                    description: _("How far the refraction eases inward from the glass edge."),
+                    type: "float",
+                    min: 1.,
+                    max: 6.5,
+                    increment: 0.1,
+                    big_increment: 0.5,
+                    digits: 2
                 },
                 falloff: {
                     name: _("Glass thickness"),
--- a/src/effects/effects.js
+++ b/src/effects/effects.js
@@ -446,16 +446,6 @@ export function get_supported_effects(_ = () => "") {
                     big_increment: 0.1,
                     digits: 2
                 },
-                webcam_gloss: {
-                    name: _("Experimental webcam gloss"),
-                    description: _("Modulates gloss from live camera brightness. This may request camera access and is experimental."),
-                    type: "boolean"
-                },
-                webcam_device: {
-                    name: _("Experimental video selector"),
-                    description: _("Camera device used by experimental webcam gloss."),
-                    type: "video_device"
-                },
                 tint: {
                     name: _("Tint"),
                     description: _("Amount of subtle milky glass tint over the blurred texture."),
--- a/src/effects/effects.js
+++ b/src/effects/effects.js
@@ -12,6 +12,7 @@ import { DerivativeEffect } from './derivative.js';
 import { RgbToHslEffect } from './rgb_to_hsl.js';
 import { HslToRgbEffect } from './hsl_to_rgb.js';
 import { LuminosityEffect } from './luminosity.js';
+import { RefractionEffect } from './refraction.js';
 
 // We do in this way because I've not found another way to store our preferences in a dictionnary
 // while calling `gettext` on it while in preferences. Not so pretty, but works.
@@ -31,6 +32,7 @@ export function get_effects_groups(_ = _ => "") {
                 "downscale",
                 "upscale",
                 "pixelize",
+                "refraction",
                 "derivative",
                 "noise",
                 "color",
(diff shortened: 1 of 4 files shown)


aunetx:
Hi! Thank you very much for the contribution :)

About the 45° corner issue (because that's the only one you mentionned?), I don't really know what it may be caused by... but that is probably inside the GLSL effect and not the pipeline right?

If that's possible you could try to rebase on the current `master` branch, as there was a big change recently! There should be nearly nothing to change for the rebasing, except for an `opacity_factor` to add in the effect, like in the other ones if I'm not mistaken.

aunetx:
By the way (I cannot test it yet), how are the performances? I guess it should be quite OK looking at the code, but just to know

treebarkbr:
> Hi! Thank you very much for the contribution :)
> 
> About the 45° corner issue (because that's the only one you mentionned?), I don't really know what it may be caused by... but that is probably inside the GLSL effect and not the pipeline right?
> 
> If that's possible you could try to rebase on the current `master` branch, as there was a big change recently! There should be nearly nothing to change for the rebasing, except for an `opacity_factor` to add in the effect, like in the other ones if I'm not mistaken.

I'll try rebasing it soon. I haven't done much to the fork lately since I've also been working on other extensions.

For the 45 degree corner issue, I think I may have found what is causing it, but I'm not completely sure yet. Since this is most likely shader-side, I may also check with winaviation and see if he has any pointers

> By the way (I cannot test it yet), how are the performances? I guess it should be quite OK looking at the code, but just to know

Performance seems pretty decent so far. I'm currently running a 3400GE with the iGPU only, and the effect is surprisingly smooth even with it applied to the top bar and Dash to Dock, plus application blur enabled.

I'll test it further and try to provide more useful performance details after that

aunetx:
Thank you! About the performances, if there is no obvious lag then that's ok, there is no need to further investigate (it is limited to on GLSL effect anyway)!

Once you have rebased and fixed the 45° corner issue, i'll be ok merging that!

aunetx:
Thanks for the update, i was actually just looking at this, what are the webcam gloss and device properties? If they are not used you maybe can remove them entirely

treebarkbr:
> Thanks for the update, i was actually just looking at this, what are the webcam gloss and device properties? If they are not used you maybe can remove them entirely

Webcam gloss was a feature that tries to shift the gloss depending on the brightest spot in the user's webcam. I thought it was on the meh side for privacy so I decided to remove it entirely. I thought of it considering Apple does the same thing with iOS/iPadOS and I thought it was pretty cool.

aunetx:
Indeed that's what I guessed! So if you're ok we can remove that, and then merge it as soon as possible (except if you prefer to wait for the 45° issue fix? because quite a big refactor of blur-my-shell is coming, so I guess you would be better off not having to rebase on it later!)

treebarkbr:
As of right now the 45° issue seems to be fixed (I forgot to port something from the original Metal shader).  I'll do some polishing first then push the changes. I think it'll be ready to merge soon.

EDIT: The issue is still there, just super subtle but I think I can fix that.

treebarkbr:
Just updating on this, there is unfortunately no fix yet (until winaviation helps me complete a deep analysis of the issue) even after rewriting the whole shader but the only fix I currently know as of now is by setting the bezel width to 20 or below, or adding a very slight Gaussian blur (preferably a radius of 3) after the Liquid Glass effect in the pipeline. Is it possible to merge this even with the issue, or not yet?

aunetx:
Thanks for trying to fix it, if the issue is subtle then of course there's no problem in merging that! I'll review one last time :)

treebarkbr:
Updating on this:
Er, it slipped past my mind that I had left my "Corner Radius" property in the Liquid Glass effect to 0 which caused the artifacts. Default should be 22, just wanting to clear things up lol.

Hazem-Samir:
when it is merged can you show how to achieve the liquid glass look ?
I am really exciting to try it

treebarkbr:
It's simple to apply, you just add the effect inside of a pipeline like this
image

Hazem-Samir:
what about the dynamic effect ?

treebarkbr:
No support for dynamic (yet) but I'll try to add support for it when I can.

Hazem-Samir:
looking forward to it

treebarkbr:
Nevermind, I read #896 and it'll support pipelines for dynamic blur once it's merged.

src/effects/refraction.glsl
```diff
@@ -0,0 +1,370 @@
...
+uniform float blur_radius;
+uniform float edge_size;
+uniform float falloff;
+uniform float corner_radius;
+uniform float rim_width;
+uniform float rgb_fringing;
+uniform float gloss;
+uniform float tint;
+uniform float shadow;
+uniform float opacity_factor;
```
krissedout: opacity_factor is declared as a uniform and the JS setter pushes it (refraction.js ~453–458), but it's never used in main(). Final alpha is only edgeOpacity:
`cogl_color_out = vec4(clamp(bgColor.rgb, 0.0, 1.0) * edgeOpacity, edgeOpacity);`

Commits were literally "Add opacity factor" / "Fix refraction opacity handling", the fix didn't make it into the shader. Either:
`cogl_color_out = vec4(..., edgeOpacity * opacity_factor);`
or drop the uniform + JS property + opacity_factor from DEFAULT_PARAMS.

src/effects/refraction.js
```diff
@@ -0,0 +1,728 @@
...
+            this.set_uniform_value('blur_direction', this.blur_direction);
+            this.set_uniform_value('private_pass', this.private_pass);
+
+            super.vfunc_paint_target(paint_node, paint_context);
+        }
+
+        vfunc_dispose() {
+            if (super.vfunc_dispose)
+                super.vfunc_dispose();
+        }
```
krissedout, line 699: Doesn't remove chained_effect from actor, clear _clip_settle_timeout_id, or disconnect theme_context. vfunc_set_actor cleans up on actor change, but dispose path can leak the chained child. Compare gaussian_blur.js which removes chained effect in vfunc_set_actor, dispose should at minimum:

GLib.Source.remove(this._clip_settle_timeout_id)
this.chained_effect?.get_actor()?.remove_effect(this.chained_effect)

src/effects/refraction.js
```diff
@@ -0,0 +1,728 @@
...
+            this._stabilize_clip_x = false;
+            this._stabilize_clip_y = false;
+            this._clip_settle_timeout_id = null;
+            this._opacity_factor = null;
+
+            utils.setup_params(this, params);
+
+            this._source = utils.get_shader_source(Shell, SHADER_FILENAME, import.meta.url);
+            if (this._source)
+                this.set_shader_source(this._source);
```
krissedout, line 213: On GNOME 48+ / snippet-only ShaderEffect, this will break however I am working on a shader_uniforms,js fix so this is non-blocking and will be fixed indirectly. (#912)

src/effects/refraction.js
```diff
@@ -0,0 +1,728 @@
...
+                0.28,
+            ),
+            'mode': GObject.ParamSpec.int(
+                `mode`,
+                `Mode`,
+                `Refraction mode`,
+                GObject.ParamFlags.READWRITE,
+                0, 1,
+                0,
+            ),
```
krissedout: mode is in DEFAULT_PARAMS, has a GObject property, setter stores _mode but never sent to the shader, not in editable_params. Remove or wire up.

src/effects/refraction.js
```diff
@@ -0,0 +1,728 @@
...
+                GObject.ParamFlags.READWRITE,
+                false,
+            ),
+            'webcam_device': GObject.ParamSpec.string(
+                `webcam_device`,
+                `Deprecated Webcam Device`,
+                `Deprecated compatibility property`,
+                GObject.ParamFlags.READWRITE,
+                '',
+            ),
```
krissedout, line 117: Deprecated compat stubs, fine to keep, but worth a comment that they're no-ops so nobody tries to use them.

src/effects/refraction.js
```diff
@@ -0,0 +1,728 @@
...
+        get strength() {
+            return this._strength;
+        }
+
+        set strength(value) {
+            if (this._strength !== value) {
+                this._strength = value;
+
+                this.set_uniform_value('strength', parseFloat(this._strength - 1e-6));
+                this.set_enabled(this.strength > 0.);
```
krissedout, line 236: Disables the entire effect (including internal blur chain) when strength is 0. User might want blur/tint with zero refraction displacement. Consider set_enabled(true) always when blur_radius > 0 or document that strength=0 disables everything.

src/effects/effects.js
```diff
@@ -378,6 +380,124 @@ export function get_supported_effects(_ = () => "") {
             }
         },
 
+        refraction: {
+            class: RefractionEffect,
+            name: _("Liquid Glass"),
+            description: _("A glossy translucent material with edge refraction, rim lighting, tint, and inner shadow."),
+            is_advanced: false,
```
krissedout, line 387: Heavy shader in the non-advanced group? Maybe intentional for discoverability, but it might spam the effects picker. Suggestion, not blocker.

krissedout (changes requested):
Code review: overall LGTM for static pipelines, the separable blur self-chain and clip stabilization are solid work.

Two things before merge:
- opacity_factor uniform is set from JS but never used in refraction.glsl main(), it looks like an incomplete fix from the opacity commits
- vfunc_dispose should clean up chained_effect and _clip_settle_timeout_id

Also heads up: this uses set_uniform_value directly, we'll want it on top of the shader uniform compatibility work from #912 

No need to wait for #896 for static blur. Happy to merge once the opacity/dispose bits are addressed. I've also added some non-blocking comments you might want to take a look at but the important ones are the aforementioned ones.

treebarkbr:
Thanks, addressed the review feedback. 

I removed opacity_factor instead of applying it because I had previously backed out the opacity behavior after it broke the refraction shader output. Also fixed dispose cleanup for the chained effect / clip settle timeout / theme context connection, removed the unused mode property, and documented the deprecated webcam compatibility stubs.

krissedout (approved):
dispose cleanup looks good, opacity dead uniform is gone. Approving and merging!

PR #907, opened 2026-06-21
Fix Dash to Panel panel blur geometry

Anduin2017:
Restore deferred handling for the Dash to Panel primary panel and recompute blur geometry from the outer DTP actor after allocation settles. This avoids partial panel blur during first initialization and makes the geometry refresh path more robust.


krissedout (approved):

krissedout:
lgtm

Issue #906, opened 2026-06-20
[Bug] Panel blur is not disabled in overview when computer boots


nvd916:
"Disable in overview" is enabled under Panel. "Panel blur" and "Background blur" use different pipelines.

This is only apparent when I set different brightness for the panel and the overview (in my case I set the brightness for the panel higher), and  only when the computer initially boots into GNOME and takes you directly to the overview.

MeguMario:
Fixed with #911

PR #911, opened 2026-06-27
Enable blur when window is close to the panel

MeguMario:
This is a small addition that add the ability to only blur the panel when a window is near the panel, basically let the panel fully transparent without blurring while it's not near anything

https://github.com/user-attachments/assets/6ef6cbd4-6519-4c82-a633-17018189b4cc

Probably not the cleanest implementation but it works.

EDIT: also include a fix to https://github.com/aunetx/blur-my-shell/issues/906


src/extension.js
```diff
@@ -448,8 +448,14 @@ export default class BlurMyShell extends Extension {
         this._settings.panel.OVERRIDE_BACKGROUND_DYNAMICALLY_changed(() => {
             if (this._settings.panel.BLUR)
                 this._panel_blur.connect_to_windows_and_overview();
+                this._panel_blur.reset();
         });
 
+        this._settings.panel.BLUR_ON_WINDOW_PROXIMITY_changed(() => {
+            if (this._settings.panel.BLUR)
+                this._panel_blur.connect_to_windows_and_overview();
+                this._panel_blur.reset();
```
krissedout, line 458: same thing here, add braces!

src/extension.js
```diff
@@ -448,8 +448,14 @@ export default class BlurMyShell extends Extension {
         this._settings.panel.OVERRIDE_BACKGROUND_DYNAMICALLY_changed(() => {
             if (this._settings.panel.BLUR)
                 this._panel_blur.connect_to_windows_and_overview();
+                this._panel_blur.reset();
```
krissedout, line 451: this always runs, its not inside the if, add braces!!

src/components/panel.js
```diff
@@ -322,15 +322,23 @@ export const PanelBlur = class PanelBlur {
...
                 );
+
                 this.connections.connect(
                     Main.overview, 'hidden', _ => this.show()
                 );
+
+                if (this.settings.panel.BLUR_ON_WINDOW_PROXIMITY && this.settings.panel.OVERRIDE_BACKGROUND_DYNAMICALLY) {
+                    this.connections.connect(
+                        Main.overview, 'hidden', _ => this.hide()
+                    );
```
krissedout: When proximity mode is on you register two hidden handlers on overview: line 327 calls show(), line 332 calls hide(). Both fire on the same signal, order-dependent and probably not intended. Consider one handler that branches on proximity state instead of stacking two connections.

src/components/panel.js
```diff
@@ -521,15 +529,27 @@ export const PanelBlur = class PanelBlur {
...
+                if (this.settings.panel.BLUR_ON_WINDOW_PROXIMITY) {
+                    if (!should_override) {
+                        this.show();
+                        panel.add_style_class_name(
+                            PANEL_STYLES[this.settings.panel.STYLE_PANEL]
+                        );
+                    }
+                    else {
+                        this.hide();
+                    };
```
krissedout: This inverts the usual dynamic behavior: !should_override -> show() + add style, should_override -> hide(). That's the opposite of the existing "disable blur when window is close" path. Worth a short comment in code explaining how this interacts with override-background-dynamically, or merging into one clear code path.

src/components/panel.js
krissedout, line 400: optional: connect_to_windows() only checks OVERRIDE_BACKGROUND_DYNAMICALLY, not BLUR_ON_WINDOW_PROXIMITY. If proximity mode is meant to drive window tracking too, this guard may need updating, otherwise proximity only works through set_should_override_panel.

src/preferences/panel.js
```diff
@@ -97,6 +107,10 @@ export const Panel = GObject.registerClass({
         );
     }
 
+    blur_proximity_option_changed() {
+        this._blur_on_window_proximity_row.set_sensitive(this.preferences.panel.OVERRIDE_BACKGROUND_DYNAMICALLY);
```
krissedout: set_sensitive only runs on OVERRIDE_BACKGROUND_DYNAMICALLY_changed, not on initial prefs load. May need a call in _init / change_blur_mode so the row starts disabled when dynamic override is off.

resources/ui/panel.ui
```diff
@@ -185,6 +185,20 @@ Recommended unless you want to customize your GNOME theme.</property>
                 </child>
               </object>
             </child>
+
+            <child>
+              <object class="AdwActionRow" id="blur_on_window_proximity_row">
+                <property name="title" translatable="yes">Enable blur when a window is near</property>
+                <property name="subtitle" translatable="yes">Enables the blur effect when a window is near the panel, and disable it otherwise.</property>
```
krissedout: Subtitle says "disable it otherwise" but doesn't mention it requires "Override background dynamically" to be on. Could add that to avoid user confusion.

schemas/org.gnome.shell.extensions.blur-my-shell.gschema.xml
```diff
@@ -282,6 +282,10 @@
             <default>false</default>
             <summary>Boolean, whether to disable blur from this component when a window is close to the panel</summary>
         </key>
+        <key type="b" name="blur-on-window-proximity">
+            <default>false</default>
+            <summary>Boolean, whether to enable blur from this component when a window is close to the panel</summary>
+        </key>
```
krissedout, line 289: These two settings are easy to confuse:

override-background-dynamically = disable blur when window is close
blur-on-window-proximity = enable blur only when window is near
Consider clarifying summaries or documenting in the UI that they're mutually exclusive modes, not two independent toggles.

krissedout (changes requested):
Nice! I've added some comments I think should be addressed. The must-fix ones before merge are 451, 457, and 330–333. The rest are polish/design.

src/extension.js
MeguMario: added, somehow I forgot (didn't throw any error so these crossed my mind)

src/extension.js
MeguMario: added, somehow I forgot (didn't throw any error so these crossed my mind lol)

schemas/org.gnome.shell.extensions.blur-my-shell.gschema.xml
MeguMario: I added the line that it required `override-background-dynamically` to be enabled, hopefully it clear enough

resources/ui/panel.ui
MeguMario: Added the clarification line.

src/preferences/panel.js
MeguMario: I added a call to this in `change_blur_mode`, hopefully this should fix that

src/components/panel.js
MeguMario: There is no modification to this on my part, and I assume this part also drive the `Disable when a window is near` (internally it is called OVERRIDE_BACKGROUND_DYNAMICALLY) so it's best to keep it general purpose for now (I know it's rather confusing).

src/components/panel.js
MeguMario: Added a comment to clariy this (though I'm rather bad at writing comment so I don't know if it's clear enough)

src/components/panel.js
MeguMario: I simplified this so now it only register one call with it calls another function now.

MeguMario:
I seperated background visibilty from `override_background_dynamic` so it should now clear the confusion
image

resources/ui/panel.ui
```diff
@@ -185,6 +185,34 @@ Recommended unless you want to customize your GNOME theme.</property>
...
+                <property name="subtitle" translatable="yes">Enables the blur effect when a window is near the panel, and disable it when it's not. Requires "Override background dynamically" to be enabled.</property>
+                <property name="activatable-widget">blur_on_window_proximity</property>
+
+                <child>
+                  <object class="GtkSwitch" id="blur_on_window_proximity">
+                    <property name="valign">center</property>
+                  </object>
+                </child>
+              </object>
+            </child>
```
krissedout: There might be a better way to do this.. Maybe you could replace the two mutually exclusive switches with one dropdown, e.g. “When window is near: background / blur / both / off”. Current UI can end up with both off, which is a footgun. This is just a suggestion, and not blocking if you fix the fallback below!

src/components/panel.js
```diff
@@ -564,20 +569,57 @@ export const PanelBlur = class PanelBlur {
...
-            this.settings.panel.OVERRIDE_BACKGROUND
-            &&
-            should_override
-        ) {
-            panel.add_style_class_name(
-                PANEL_STYLES[this.settings.panel.STYLE_PANEL]
-            );
-        }
+        if (this.settings.panel.OVERRIDE_BACKGROUND) {
+            if (this.settings.panel.OVERRIDE_BACKGROUND_DYNAMICALLY) {
```
krissedout, line 576: When dynamic override is on but both proximity toggles are off, neither inner branch runs and we lose the old if (should_override) add_style_class_name(...) behavior. Add a fallback for that case.

src/components/panel.js
```diff
@@ -365,23 +369,24 @@ export const PanelBlur = class PanelBlur {
                 this.connections.connect(
                     Main.overview, 'showing', _ => this.hide()
                 );
+                
                 this.connections.connect(
-                    Main.overview, 'hidden', _ => this.show()
+                    Main.overview, 'hidden', _ => this.panel_hide_blur_dynamically()
```
krissedout: panel_hide_blur_dynamically() is only called with no args, so should_override is always undefined and the inner if (should_override) never runs. Either pass real state or call update_visibility() here.

src/components/panel.js
```diff
@@ -365,23 +369,24 @@ export const PanelBlur = class PanelBlur {
...
+                
                 this.connections.connect(
                     appDisplay, 'show', _ => this.hide()
                 );
+
                 this.connections.connect(
                     appDisplay, 'hide', _ => this.show()
                 );
                 this.connections.connect(
                     Main.overview, 'hidden', _ => this.show()
```
krissedout: Unconditional show() breaks blur-on-proximity: after overview, blur can show when no window is near. Use update_visibility() instead.

src/components/panel.js
```diff
@@ -564,20 +569,57 @@ export const PanelBlur = class PanelBlur {
...
     }
 
+    panel_hide_blur_startup(){
+        if (this._first_boot) {
+            this.hide();
+            this._first_boot = false
+        }
+    }
+
+    panel_hide_blur_dynamically(should_override){
```
krissedout: Dead parameter, only call site is line 374 with no args. Remove or wire up.

krissedout (changes requested):
Thanks for the fixes! extension.js braces look good. I've done another review for suggestions.

resources/ui/panel.ui
MeguMario: I replaced this with a dropdown as you said, though the later "both" and "off" options are not really needed because for "off" you can just disable dynamic background override, and "both" really defeat the point of Panel blur as you can just get this functionally from other extension that specialize in this.

src/components/panel.js
MeguMario: removed

src/components/panel.js
MeguMario: Fallback is not needed as I switched to dropdown now.

src/components/panel.js
MeguMario: this I will keep the same as `should_override` should be already set by the logic that drive `override-background-dynamically` before.

MeguMario: Though I think I will switch to `update_visibility()` as it doesn't seem to change all that much.

src/components/panel.js
MeguMario: Switched to `update_visibility() `.

MeguMario:
It's dropdown menu time
image

src/components/panel.js
```diff
@@ -563,20 +572,60 @@ export const PanelBlur = class PanelBlur {
...
+        if (this.settings.panel.OVERRIDE_BACKGROUND) {
+            if (this.settings.panel.OVERRIDE_BACKGROUND_DYNAMICALLY) {
+                // This is an invert of the above behavior, 
+                // Blur and all styling is hidden when "should_override" is true. 
+                if (this.settings.panel.OVERRIDE_BACKGROUND_DYNAMICALLY_MODE == 0) {
+                    panel.add_style_class_name(
+                        PANEL_STYLES[this.settings.panel.STYLE_PANEL]
+                    );
+                    if (!should_override) {
+                        this.show();
```
krissedout: set_should_override_panel runs once per panel inside update_visibility(), but this.show() / this.hide() touch every entry in actors_list (see 637–646). On multi-monitor, the last panel processed wins, e.g. monitor A has a nearby window (should show blur) and monitor B doesn’t; one of them will be wrong.

Fix: use per-panel actors here:
```
if (!should_override)
    actors.widgets.background.show();
else
    actors.widgets.background.hide();
```

krissedout (changes requested):
Dropdown + overview/hidetopbar paths look good, previous blockers are addressed.

Still blocking: in mode 0 (OVERRIDE_BACKGROUND_DYNAMICALLY_MODE == 0), set_should_override_panel must not call this.show() / this.hide() as those are global. Use actors.widgets.background.show() / .hide() per panel instead. Same class of bug on destroy_blur calling into that path.

Single-monitor is probably fine; multi-monitor / multi-panel will be wrong until fixed.

Also the subtitle grammar is a bit messy and long, try to clean it up but that's cosmetic only!

src/components/panel.js
MeguMario: I already applied the fix as you said. Thanks for the suggestion!

krissedout (approved):
Alright LGTM now! I'm approving and you can merge if you feel like it's ready and don't have anything to add.

MeguMario:
> Alright LGTM now! I'm approving and you can merge if you feel like it's ready and don't have anything to add.

Thanks a lot! I don't think there are more to add so I think I'm going to merge this. 

On a side note, the current dynamic background overriding implementation is still rather shaky on multi monitor case (especially one with the panel on top but primary monitor is on the bottom) and having allocation log spam, so a fix for that is going to be needed in the future.

PR #912, opened 2026-06-28
Fix shader uniform handling for GNOME 48+ compatibility

krissedout:
## Summary

- Adds `shader_uniforms.js` to centralize Clutter effect uniform uploads and avoid GNOME 48+ breakage from direct `set_uniform_value` usage.
- Updates all effect GLSL/JS pairs to use consistent uniform names and deferred upload via `upload_uniforms`.
- Extracted from the shared blur surfaces refactor (`dc49123`) without the larger architectural changes or the experimental Cogl snippet rewrite.

### DIFF
--- a/src/conveniences/pipeline.js
+++ b/src/conveniences/pipeline.js
@@ -2,6 +2,7 @@ import St from 'gi://St';
 import Meta from 'gi://Meta';
 import * as Main from 'resource:///org/gnome/shell/ui/main.js';
 import * as Background from 'resource:///org/gnome/shell/ui/background.js';
+import * as uniforms from './shader_uniforms.js';
 
 /// A `Pipeline` object is a handy way to manage the effects attached to an actor. It only manages
 /// one actor at a time (so blurring multiple widgets will need multiple `Pipeline`), and is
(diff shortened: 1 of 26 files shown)


aunetx (approved):
I could not test it, but the code appears right to me! Although a bit verbose for the GLSL types in my opinion but I know you like it like that haha

So looks good to me if you/someone else have tested it!

MeguMario:
While this is probably intended for the refactor, I have to point out that while the corner effect is still working, the option to do so is missing so if you haven't added that before then you can't add that within the prefs menu

With the patch
image

Without the patch
Screenshot From 2026-07-02 22-13-14

Other than that, most effects seems to work except for Pixelize, which stopped working (probably ever since popup blur merged)

krissedout:
@MeguMario I forgot to readd the corner effect in the effects.js after splitting this out of the refactor! I've re-added it, also I didn't port the build pixelize effect from the refactor so I've ported that into here too, please retest when you can!

MeguMario:
@krissedout ok it's seems good to go now, thanks for the fixed!

PR #913, opened 2026-06-29
fix(panel): apply shadow artifact repaint hack to static blur

### LINKED ISSUE #886: [Bug] Panel blur breaks shortly after the extension loads
Fedora 43 (Workstation Edition), GNOME 49, Blur my Shell version 72

Every session I start, the top bar is blurred probably for a couple minutes, then it starts degrading -- notifications leave a non-blurred trail that flickers rapidly, hovering over dynamic elements in the bar (system vitals from Vitals extension, network meter from Net Speed extension, system clock etc.) removes the blur behind them, but also sometimes brings it back? Sometimes value updates bring back blur only for it to disappear completely, or just behind the text a couple updates later.

My current Panel blur settings are:
- Static blur
- Default pipeline -- just a Gauss blur
- Force light text disabled
- Override background enabled (Transparent background style, disable when window disabled, disable in overview enabled)

I've enabled Dynamic blur just for the sake of testing, and it seems to be more stable, I'll report back if a similar issue occurs.

Thetwobraincells:
Resolves #886 where static blur on the panel degrades, flickers, 
or leaves non-blurred trails under dynamic UI elements (e.g., Vitals, 
Net Speed, clock updates). This is caused by a known GNOME Shell 
compositor bug where \`Shell.BlurEffect\` fails to repaint properly 
under shadows.

Previously, the \`paint_signals\` workaround (\`HACKS_LEVEL === 1\`) was 
exclusively applied to the dynamic blur pipeline. This commit hoists 
the hack so it applies to both static and dynamic blur configurations.

Implementation details:
- Introduces a duck-typed proxy object (\`dynamic_target\`) to intercept 
  \`queue_repaint\` calls.
- Dynamically resolves the active blur effect at runtime. For the static 
  \`Pipeline\`, it iterates safely through the effects array to find the 
  blur effect; for \`DummyPipeline\`, it falls back to the singular effect.
- Ensures the hack survives pipeline rebuilds (e.g., when a user changes 
  blur settings) without causing crashes or memory leaks."

### DIFF
--- a/src/components/panel.js
+++ b/src/components/panel.js
@@ -200,9 +200,11 @@ export const PanelBlur = class PanelBlur {
 
         let background, bg_manager;
         let static_blur = this.settings.panel.STATIC_BLUR;
+        let pipeline; // Hoist the pipeline variable so our proxy can access it
+
         if (static_blur) {
             let bg_manager_list = [];
-            const pipeline = new Pipeline(
+            pipeline = new Pipeline(
                 this.effects_manager,
                 global.blur_my_shell._pipelines_manager,
                 this.settings.panel.PIPELINE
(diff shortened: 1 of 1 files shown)


src/components/panel.js
```diff
@@ -214,32 +216,56 @@ export const PanelBlur = class PanelBlur {
...
+
+                // Proxy object to dynamically resolve the active blur effect.
+                // This ensures repaints continue safely even if pipeline effects are rebuilt.
+                let dynamic_target = {
+                    queue_repaint: () => {
+                        let eff = null;
+                        if (static_blur && pipeline && pipeline.effects) {
+                            // Find the blur effect in the static pipeline array
+                            for (let i = 0; i < pipeline.effects.length; i++) {
+                                if (pipeline.effects[i]._bms_effect_type && pipeline.effects[i]._bms_effect_type.includes('blur')) {
```
krissedout: This only finds effects whose _bms_effect_type contains 'blur'. #893 got merged which means a panel pipeline can be refraction / Liquid Glass with no blur-type effect. You’ll miss it or fall back to effects[0], which may be the wrong effect in the chain. Please resolve the active pipeline effect properly, e.g. use bg_manager._bms_pipeline and repaint the last effect in the chain, or all pipeline effects, not blur-by-substring.

src/components/panel.js
```diff
@@ -214,32 +216,56 @@ export const PanelBlur = class PanelBlur {
...
+                                    eff = pipeline.effects[i];
+                                    break;
+                                }
+                            }
+                            if (!eff) eff = pipeline.effects[0]; // Fallback to first effect
```
krissedout: Fallback to effects[0] is arbitrary if the first effect isn’t the one that needs repainting. Same fix as above.

krissedout (changes requested):
Good direction for #886, but the blur-only effect lookup won’t survive refraction/Liquid Glass pipelines (#893). You need a pipeline-agnostic way to pick which effect(s) to queue_repaint().

src/components/panel.js
Thetwobraincells: I've updated the PR to drop the substring matching entirely.
It now dynamically resolves the active pipeline via bg_manager._bms_pipeline and iterates through to call queue_repaint() on all effects in the active chain. This makes it completely agnostic to the current pipeline configuration and avoids the arbitrary effects[0] fallback.
Let me know if this looks good to you.

src/components/panel.js
Thetwobraincells: Fixed in above.

krissedout (approved):
Static repaint hack + active_pipeline + repainting all static effects looks correct for #886 and refraction pipelines. Approving.

PR #916, opened 2026-07-01
fix(panel): track inner panel for Dash to Panel dynamic length

### LINKED ISSUE #873: Support Dash-to-Panel's dynamic panel length
Note: no LLM ("AI") was used in any way, directly or indirectly. All the code digging, implementation and issue write-up was done by a brain controlling a flesh robot (aka human).

### Abstract
**I'd like for Blur My Shell to support Dash to Panel (DtP)'s dynamic panel length (dock mode). I provide a rudimentary implementation that works for dynamic blur on center-anchored dynamic length panels, in hopes that it helps with proper first-party implementation.**

While Blur My Shell has good support for panels created by Dash to Panel, it assumes that the size of the panel is set by the DtP extension settings and does not change until the next settings modification. However, this assumption does not hold after the introduction of dynamic panel length (dock mode) since [DtP v66](https://github.com/home-sweet-gnome/dash-to-panel/releases/tag/v66), circa March 2025. 

### Issue
When dynamic blur is applied on DtP panels, the blur area is incorrect, and does not update its size as the panel length changes:

Image

Image

### Half-baked Implementation
I will not be opening a PR since I know that my implementation is nowhere near production ready, and I regrettably do not have time to make it so. Basically, I create a new panel updating function and connect it to the DtP panel's notify::size signal. 

**Caveats:** besides the hard-coded nature of things, ~~this causes the spamming of `Can't update stage views actor bms-panel-blurred-widget [StWidget] is on because it needs an allocation.` and `Can't update stage views actor bms-panel-backgroundgroup [MetaBackgroundGroup] is on because it needs an allocation.`~~ Probably because I don't completely understand how BMS works.  (Edit: I still don't really understand GJS, but I fixed this spamming issue by using GLib.add_idle just like how you implemented `blur_dtp_panels()` ).

Here's the git diff:

```
diff --git a/src/components/panel.js b/src/components/panel.js
index bd39166..7116d86 100644
--- a/src/components/panel.js
+++ b/src/components/panel.js
@@ -247,28 +247,40 @@ export const PanelBlur = class PanelBlur {
 
         // connect to panel, panel_box and its parent position or size change
         // this should fire update_size every time one of its params change
-        this.connections.connect(
-            panel,
-            'notify::position',
-            _ => this.update_size(actors)
-        );
-        this.connections.connect(
-            panel_box,
-            ['notify::size', 'notify::position'],
-            _ => this.update_size(actors)
-        );
-        this.connections.connect(
-            panel_box.get_parent(),
-            'notify::position',
-            _ => this.update_size(actors)
-        );
-
+        if(!is_dtp_panel){        
+        
+            this.connections.connect(
+                panel,
+                'notify::position',
+                _ => this.update_size(actors)
+            );
+            this.connections.connect(
+                panel_box,
+                ['notify::size', 'notify::position'],
+                _ => this.update_size(actors)
+            );
+            this.connections.connect(
+                panel_box.get_parent(),
+                'notify::position',
+                _ => this.update_size(actors)
+            );
+        }
+        
         // connect to the panel getting destroyed
         this.connections.connect(
             panel,
             'destroy',
             _ => this.destroy_blur(actors, true)
         );
+
+        if(is_dtp_panel){
+            this.dtp_update_size(actors); //call once after init
+            this.connections.connect(
+                panel,
+                'notify::size',
+                _ => this.dtp_update_size(actors)
+            );
+        }
     }
 
     update_size(actors) {
@@ -305,6 +317,43 @@ export const PanelBlur = class PanelBlur {
         actors.monitor = Main.layoutManager.findMonitorForActor(panel);
     }
 
+    dtp_update_size(actors) {
+        GLib.idle_add(GLib.PRIORITY_DEFAULT_IDLE, () => {
+            let panel = actors.widgets.panel;
+            let panel_box = actors.widgets.panel_box;
+            let background = actors.widgets.background;
+            let [width, height] = panel_box.get_size();
+
+            let monitor = Main.layoutManager.findMonitorForActor(panel);
+            //not sure if these are needed here tbh
+            //while (!monitor){
+            //    monitor = Main.layoutManager.findMonitorForActor(panel);
+            //}
+        
+            // if static blur, need to clip the background
+            if (actors.static_blur) {  
+                // an alternative to panel.get_transformed_position, because it
+                // sometimes yields NaN (probably when the actor is not fully
+                // positionned yet)
+                let [p_x, p_y] = panel_box.get_position();
+                let [p_p_x, p_p_y] = panel_box.get_parent().get_position();
+                let x = p_x + p_p_x - monitor.x + (width - panel.width) / 2;
+                let y = p_y + p_p_y - monitor.y + (height - panel.height) / 2;
+
+                background.set_clip(x, y, panel.width, panel.height);
+                background.x = (width - panel.width) / 2 - x;
+                background.y = .5 + (height - panel.height) / 2 - y;
+            } else {
+                background.x = panel.x + (monitor.width - panel.width)/2 - 4;//manual fix for my monitor
+                background.y = panel.y;
+                background.width = panel.width;
+                background.height = panel.height;
+            }
+
+            return GLib.SOURCE_REMOVE;
+        });
+    }
+
     /// Connect when overview if opened/closed to hide/show the blur accordingly
     ///
     /// If HIDETOPBAR is set, we need just to hide the blur when showing appgrid
```

Thetwobraincells:
Fixes #873 to ensure the blur pipeline correctly maps to the inner panel's boundaries in real-time.

Targeting the inner panel: Reassigned the geometry_actor to the inner panel instead of the wrapper when a Dash to Panel structure is detected.

Adding wrapper listeners: Injected geometry signal listeners (notify::allocation, notify::size, notify::position) into the wrapper to accurately track horizontal sliding and structural shifts.

Updating coordinate math: Adjusted the update_size() calculations for both static and dynamic blur styles to correctly sum the wrapper's absolute position with the inner panel's relative coordinates.

### DIFF
--- a/src/components/panel.js
+++ b/src/components/panel.js
@@ -186,11 +186,13 @@ export const PanelBlur = class PanelBlur {
     /// Blur a panel
     blur_panel(panel) {
         let geometry_actor = panel;
+        let wrapper = null;
         let panel_box = panel.get_parent();
         let is_dtp_panel = false;
         if (!panel_box.name) {
             is_dtp_panel = true;
-            geometry_actor = panel; // track inner panel instead of the wrapper
+            wrapper = panel_box;
+            geometry_actor = panel; // Track the inner panel dynamically
             panel_box = panel_box.get_parent();
         }
 
@@ -279,6 +281,7 @@ export const PanelBlur = class PanelBlur {
         let actors = {
             widgets: {
                 panel,
+                wrapper,
                 panel_box,
                 background,
                 background_group,
@@ -301,6 +304,14 @@ export const PanelBlur = class PanelBlur {
             ['notify::allocation', 'notify::size', 'notify::position'],
             _ => this.update_size(actors)
         );
+
+        if (wrapper) {
+            this.connections.connect(
+                wrapper,
+                ['notify::allocation', 'notify::size', 'notify::position'],
+                _ => this.update_size(actors)
+            );
+        }
         this.connections.connect(
             panel_box,
             ['notify::size', 'notify::position'],
@@ -339,6 +350,7 @@ export const PanelBlur = class PanelBlur {
     update_size(actors) {
         let geometry_actor = actors.widgets.geometry_actor;
         let panel_box = actors.widgets.panel_box;
+        let wrapper = actors.widgets.wrapper;
         let background = actors.widgets.background;
         let [width, height] = panel_box.get_size();
         let [geometry_width, geometry_height] = geometry_actor.get_size();
@@ -358,12 +370,16 @@ export const PanelBlur = class PanelBlur {
 
             let [p_x, p_y] = panel_box.get_position();
             let [p_p_x, p_p_y] = panel_box.get_parent().get_position();
-            let [g_x, g_y] = geometry_actor.get_position();
 
+            // updated coordinates for static blur
+            let g_x, g_y;
             if (actors.is_dtp_panel) {
-                let [w_x, w_y] = geometry_actor.get_parent().get_position();
-                g_x += w_x;
-                g_y += w_y;
+                let [w_x, w_y] = wrapper.get_position();
+                let [pan_x, pan_y] = geometry_actor.get_position();
+                g_x = w_x + pan_x;
+                g_y = w_y + pan_y;
+            } else {
+                [g_x, g_y] = geometry_actor.get_position();
             }
 
             let x = p_x + p_p_x - monitor.x + g_x;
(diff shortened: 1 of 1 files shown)


krissedout:
lgtm, but a minor question before approving, could you check your logs and see if you have the allocation warnings, if so you could use queue_update_size() for consistency with the initial defer path

Thetwobraincells:
> lgtm, but a minor question before approving, could you check your logs and see if you have the allocation warnings, if so you could use queue_update_size() for consistency with the initial defer path

I monitored the GNOME Shell logs while resizing and interacting with the dynamic dock. In my test, the logs were actually completely clean of any `Clutter-CRITICAL` allocation or layout cycle warnings. 
However, I have updated all the geometry and allocation signal connections to use `queue_update_size()` to safely defer the math to the idle loop. 
The updated commit has been pushed. Thanks for pointing that out.

krissedout (approved):
Alright then! Approving.

nonstandardusr:
Nice!

PR #918, opened 2026-07-03
Remove corner_radius from DEFAULT_PARAMS

MeguMario:
Remove `corner_radius` from `DEFAULT_PARAMS` for Dynamic blur to fix a regression from https://github.com/aunetx/blur-my-shell/pull/912 (a very dirty one at that) that made the corner radius use the default param on startup instead of using the one from user option

This value doesn't seem to do anything as everything else use the value from user option anyway.


aunetx (commented):
LGTM :)

aunetx (approved):

PR #919, opened 2026-07-04
fix(applications): correct blur widget name check in opacity guard

### LINKED ISSUE #909: Application blur backdrop turns transparent: set_window_opacity checks the wrong actor name ("blur-actor" vs "bms-application-blurred-widget")
## Description

With application blur enabled, the blurred **background actor itself** is made semi-transparent together with the window. The blur backdrop ends up washed out / too transparent, and on GTK4 apps it can look non-uniform and intermittent — most visible on **Nautilus**, during its opacity transitions, focus changes or resize.

## Root cause

In `src/components/applications.js`, `set_window_opacity()` applies the opacity to every child of the window actor **except** the one named `"blur-actor"`:

```js
set_window_opacity(window_actor, opacity) {
    window_actor?.get_children().forEach(child => {
        if (child.name !== "blur-actor" && child.opacity != opacity)
            child.opacity = opacity;
    });
}
```

But the blurred widget is created with the name **`"bms-application-blurred-widget"`**, not `"blur-actor"`:

- `create_blur_effect()` passes `'bms-application-blurred-widget'` as the widget name;
- `DummyPipeline.create_background_with_effect()` (`conveniences/dummy_pipeline.js`) does `this.actor = new St.Widget({ name: widget_name })` and inserts it as a **direct child** of the window actor (`background_group.insert_child_at_index(this.actor, 0)`), returning it as the `blur_actor`.

So `child.name !== "blur-actor"` is always true for the blur widget → the backdrop receives the reduced window `OPACITY` → washed-out blur. It looks like a rename that wasn't propagated to this guard.

The effect is amplified by the `notify::opacity` handler (`applications.js`, in `create_blur_effect`) which re-applies the opacity on every child on each change; combined with GTK4 apps animating their own opacity (Nautilus), the transparency is applied unevenly and intermittently.

## Proposed fix

Exclude the actual blur widget name (or, better, compare against the `blur_actor` reference instead of a name):

```js
if (child.name !== "blur-actor" && child.name !== "bms-application-blurred-widget" && child.opacity != opacity)
```

I've been running this one-line change locally and the backdrop stays opaque as intended.

## Environment

- Blur my Shell **v72** (same code is present on current `master`, line 522)
- GNOME Shell 50.2, Fedora, Wayland

Thetwobraincells:
fixes #909 

The blur backdrop was turning transparent during window animations because `set_window_opacity` was checking for the old "blur-actor" name instead of "bms-application-blurred-widget". Added the correct name to the exclusion guard to keep the backdrop fully opaque.

### DIFF
--- a/src/components/applications.js
+++ b/src/components/applications.js
@@ -518,9 +518,14 @@ export const ApplicationsBlur = class ApplicationsBlur {
 
     /// Set the opacity of the window actor that sits on top of the blur effect.
     set_window_opacity(window_actor, opacity) {
+        // Define known blur actor names. This makes it easy to update if names change again.
+        const BLUR_ACTOR_NAMES = new Set(["blur-actor", "bms-application-blurred-widget"]);
+
         window_actor?.get_children().forEach(child => {
-            if (child.name !== "blur-actor" && child.opacity != opacity)
+            // Check against the Set and the opacity
+            if (!BLUR_ACTOR_NAMES.has(child.name) && child.opacity != opacity) {
                 child.opacity = opacity;
+            }
         });
     }
 


krissedout (approved):
LGTM!

PR #896, opened 2026-06-05
Refactor blur surfaces and preferences

krissedout:
This refactors Blur my Shell around shared surface blur primitives instead of each component owning its own blur lifecycle.

- Add shared static/dynamic blur surface infrastructure
- Support pipelines on dynamic blur using live source snapshots, while keeping the native dynamic blur fast path when possible
- Move panel, Dash to Dock, applications, popups, and related surfaces onto the shared runtime
- Consolidate preferences into a Surfaces page with a global blur profile and per-surface overrides
- Move corner radius handling out of the default rounded pipeline and into surface settings
- Remove the old rounded blur helper/library flow

As I haven't end to end tested this yet, I'm opening this PR as a draft for anyone that wants to test it and give feedback, it could likely need some fixes or changes.

PR #924, opened 2026-07-07
Fold Dash blur into Popup blur

### LINKED ISSUE #880: Dhruva Integration
I have switched to using a new dock called [Dhruva](https://github.com/NarkAgni/dhruva).

However, background blur (for frosted glass) does not work with this dock.  It appears to only be designed for dash to dock.

It would be nice if there was support for Dhruva as well.

MeguMario:
This is an experiemental change that fold dash blur into popup blur.

I think the architecture of popup blur allows for much more generic support for dock blurring, so for that, this PR simply add `dash-background` and `plank-like-dock-bg` classes to `POPUP_TARGET_STYLE_CLASSES`  and add seperate corner radius control for the dash(es) inside popup blur settings.

With dash to dock
image
image

With dhruva

https://github.com/user-attachments/assets/ffc5d78d-1530-48e8-aca3-8a6b0cce8634

https://github.com/user-attachments/assets/c144058f-a854-4130-889c-5cb380763a1b

With Dock!ng
image
image

### Todo
- [x] Reimplement some of the feature from the current Dash blur
- [ ] Test for multi-monitor scenario


src/components/popup/blur_surface.js
```diff
@@ -230,6 +243,66 @@ export const PopupBlurSurface = class PopupBlurSurface {
...
+
+            if (!is_dash_surface())
+                return;
+
+            this._first_boot = false;
+
+            this.hide_surface_overview();
+
+            // We show the overview to workaround the issue where the dash background is not visible on option enable while the overview is hidden.
+            Main.overview.show();
```
krissedout, line 264: Calling Main.overview.show() here forces Activities open whenever a dash surface is created with “Disable Dash blurring in overview” enabled (extension enable / popup.reset() / toggling the setting). The old dash blur only hid/showed on overview signals. Please drop this workaround and fix the dash visibility issue without changing the user’s session state, multi-monitor will also hit this once per dock.

src/components/popup/blur_surface.js
```diff
@@ -230,6 +243,66 @@ export const PopupBlurSurface = class PopupBlurSurface {
...
+        }
+    }
+
+    connect_to_overview() {
```
krissedout, line 268: connect_to_overview() runs for every popup surface when the setting is on, and destroy() never disconnects these handlers, only a full popup disable()/reset() does. That leaks showing/hidden listeners as menus open/close and as docks recreate. Prefer one overview connection on PopupBlur, or connect only for dash surfaces and disconnect in destroy().

src/styles/popup/base.css
```diff
@@ -192,3 +192,15 @@
 .bms-popup-background-dark .popup-sub-menu .popup-menu-section:last-child .popup-menu-item:last-child {
     border-radius: 0 0 12px 12px;
 }
+
+.bms-popup-background-transparent .dash-background {
```
krissedout, line 196: Blur targets include plank-like-dock-bg, but background override CSS only covers .dash-background. Dock!ng-style docks can keep an opaque background over the blur, please add matching .plank-like-dock-bg rules (or share a selector).

schemas/org.gnome.shell.extensions.blur-my-shell.gschema.xml
```diff
@@ -667,6 +601,11 @@
             <default>3</default>
             <summary>Enum to select the style of popup surfaces (0 transparent, 1 light, 2 dark, 3 automatic)</summary>
         </key>
+        <!-- UNBLUR IN OVERVIEW DASH -->
+        <key type="b" name="unblur-in-overview-dash">
```
krissedout, line 605: unblur-in-overview-dash (and dash corner radius) replace the old dash-to-dock keys, but there’s no migration from the removed schema. Existing users will lose blur-on/off, pipeline/static mode, style, corner radius, and unblur-in-overview. Worth migrating at least the important ones, or documenting the break clearly.

src/components/popup/index.js
```diff
@@ -276,7 +276,7 @@ export const PopupBlur = class PopupBlur {
             if (parent === global.window_group)
                 return { parent: actor, sibling: child ?? actor.get_last_child?.() ?? null };
 
-            child = actor;
+            child = actor;PopupBlurSurface
```
krissedout, line 279: Accidental leftover: should be child = actor; PopupBlurSurface here is a no-op.

krissedout (changes requested):
Really nice!! This will likely help take the burden off of #896 later so thank you for that! Before merge I’d want the Overview show() workaround gone, and overview listeners scoped/cleaned up properly. Also worth a settings migration note for users coming off the old dash-to-dock schema (multi-monitor still unchecked on the PR).

Also styles/dash.css was deleted and the rules moved into popup/base.css, but this import remains. Please remove it so theme load doesn’t depend on a missing file.

And we could start using the canary approach now likely because this is a big overhaul, but it might not be needed yet, let's just see after the fixes!

MeguMario:
Some changes already made per review but I think it's best to wait for #932 to merge so that we can rebase this PR before pushing

krissedout:
This has been superseded by #976, which moves Dash and the other blur implementations onto the shared backend. I think we can close this in favor of that.

MeguMario:
> This has been superseded by #976, which moves Dash and the other blur implementations onto the shared backend. I think we can close this in favor of that.

I think so too, let me close this real quick

Side note that we should probably expand the current dash blur implementation to other dash implementation as well (which this PR currently do)

krissedout:
> > This has been superseded by #976, which moves Dash and the other blur implementations onto the shared backend. I think we can close this in favor of that.
> 
> I think so too, let me close this real quick
> 
> Side note that we should probably expand the current dash blur implementation to other dash implementation as well (which this PR currently do)

I added the Dhruva integration from here into there too!

PR #925, opened 2026-07-07
Fix Liquid Glass fade to black in Popup blur

MeguMario:
This PR fix the issue where Liquid Glass texture effect fade to black in Popup Blur instead of properly fade out like other effect

https://github.com/user-attachments/assets/ace19b79-d202-4918-89c2-4a08f2b85878

### DIFF
--- a/src/components/popup/static_actor.js
+++ b/src/components/popup/static_actor.js
@@ -68,6 +68,7 @@ export const PopupBlurStaticActor = class PopupBlurStaticActor {
                     upscale: () => this.get_texture_effect_overrides(),
                     pixelize: () => this.get_texture_effect_overrides(),
                     derivative: () => this.get_texture_effect_overrides(),
+                    refraction: () => this.get_texture_effect_overrides(),
                     color: params => this.get_color_effect_overrides(params),
                     luminosity: () => this.get_luminosity_effect_overrides(),
                     noise: params => this.get_noise_effect_overrides(params),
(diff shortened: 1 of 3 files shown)


src/effects/refraction.js
```diff
@@ -480,6 +490,17 @@ export const RefractionEffect = utils.IS_IN_PREFERENCES
             this.update_scaled_uniforms();
         }
 
+        get opacity_factor() {
+            return this._opacity_factor;
+        }
+
+        set opacity_factor(value) {
```
krissedout, line 497: nit: other refraction setters propagate to chained_effect (see edge_size, blur_radius, etc.). Worth doing the same here:
```
if (this.chained_effect)
    this.chained_effect.opacity_factor = value;
```
Pass 1 doesn't use the uniform in the shader today, so probably not blocking but it keeps behavior consistent if that path changes later.

src/components/popup/static_actor.js
```diff
@@ -222,7 +223,7 @@ export const PopupBlurStaticActor = class PopupBlurStaticActor {
 
         return overrides;
     }
-
+
```
krissedout: why is there a stray newline?

krissedout (approved):
lgtm, minor nits but it's fine as-is if you tested it!

src/components/popup/static_actor.js
MeguMario: Probably because I press my tab key on accident haha. I deleted it now though

src/effects/refraction.js
MeguMario: Added that!

Issue #865, opened 2026-04-24
Multi Monitor: Disable when a window is near


DasJott:
I use my Laptop monitor as primary, therefore with a panel on it.
Above that display is another monitor.

Everytime a window exists on that laptop monitor the panel is not transparent/blurred when the option "Disable when a window is near" is checked. No matter if the window is even near the panel.

So in this monitor setup (anotehr monitor above the primary one) it seems that this extension recognizes every window as "near to panel" if it is below the panel.
If I drag that window up to the other monitor, the panel is transparent again.

PR #926, opened 2026-07-07
fix(panel): multi-monitor vertical layout proximity detection

Thetwobraincells:
When a secondary monitor is positioned physically above the primary monitor, the primary monitor's Y-origin is pushed down the coordinate plane (e.g., Y = 1080). The update_visibility logic in panel.js hardcoded panel_top === 0 to detect top panels. Consequently, top panels on lower monitors were evaluated as panel_top > 0 (treating them as bottom panels). This caused the "Disable when a window is near" feature to trigger false positive intersections for almost all windows on the primary display.

Fix: Replaced the hardcoded 0 with actors.monitor.y to dynamically respect the panel's position relative to its specific monitor's bounding box, resolving Issue #865 without breaking side-by-side or single monitor configurations.

### DIFF
--- a/src/components/panel.js
+++ b/src/components/panel.js
@@ -603,10 +603,10 @@ export const PanelBlur = class PanelBlur {
                     // if so, and if in the same monitor, then it overlaps
                     if (same_monitor
                         &&
-                        // check if panel is on top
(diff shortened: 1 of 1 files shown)


MeguMario (approved):
LGTM!

PR #927, opened 2026-07-10
Gradient for Blur on Window proximity

MeguMario:
Add gradient for Blur on Window Proximity for better visibility

image

Without gradient it looks like this on bright wallpaper
Screenshot From 2026-07-10 17-43-40

Can be enable / disable in Settings (Default to Off)

image

Still figuring out how to make the panel stop flashing black for a few seconds when changing style class, if anyone can help me find the solution then I appreciate it!

### DIFF
--- a/resources/ui/panel.ui
+++ b/resources/ui/panel.ui
@@ -202,6 +202,35 @@ Background on Window Proximity: Solid color background when a window is near the
               </object>
             </child>
 
+            <child>
+              <object class="AdwActionRow" id="gradient_panel_row">
+                <property name="title" translatable="yes">Gradient background</property>
+                <property name="subtitle" translatable="yes">Apply a gradient effect to the panel background.</property>
+                <property name="activatable-widget">gradient_panel</property>
+
+                <child>
+                  <object class="GtkSwitch" id="gradient_panel">
+                    <property name="valign">center</property>
+                  </object>
+                </child>
+              </object>
+            </child>
+
+            <child>
+              <object class="AdwActionRow" id="gradient_panel_mode_row">
+                <property name="title" translatable="yes">Gradient direction</property>
+                <property name="subtitle" translatable="yes">Change the direction of the gradient background.</property>
+                <property name="activatable-widget">gradient_panel_mode</property>
+
+                <child>
+                  <object class="GtkDropDown" id="gradient_panel_mode">
+                    <property name="valign">center</property>
+                    <property name="model">gradient_panel_mode_model</property>
+                  </object>
+                </child>
+              </object>
+            </child>
+
           </object>
         </child>
 
@@ -273,6 +302,13 @@ Background on Window Proximity: Solid color background when a window is near the
     <items>
       <item translatable="yes">Blur on Window Proximity</item>
       <item translatable="yes">Background on Window Proximity</item>
+    </items>
+  </object>
+
+  <object class="GtkStringList" id="gradient_panel_mode_model">
+    <items>
+      <item translatable="yes">Top</item>
+      <item translatable="yes">Bottom</item>
     </items>
   </object>
 
--- a/src/components/panel.js
+++ b/src/components/panel.js
@@ -16,6 +16,11 @@ const PANEL_STYLES = [
     "contrasted-panel"
 ];
 
+const GRADIENT_PANEL_STYLES = [
+    "gradient-panel",
+    "gradient-panel-reverse"
+];
+
 // global listener, so we don't miss the panel destruction event
 let isMainPanelAlive = true;
 Main.panel.connect('destroy', () => isMainPanelAlive = false);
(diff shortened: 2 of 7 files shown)


src/components/panel.js
```diff
@@ -672,6 +675,41 @@ export const PanelBlur = class PanelBlur {
...
+                    panel.add_style_class_name(
+                        GRADIENT_PANEL_STYLES[this.settings.panel.GRADIENT_PANEL_MODE]
+                    );
+                }
+            }
+        }
+        else {
+            panel.add_style_class_name(
+                PANEL_STYLES[0]
+            );
```
krissedout: When GRADIENT_PANEL is off (default), this forces transparent-panel on hide.

On master, mode 0 kept STYLE_PANEL and only hid the blur actor. That means existing “Blur on Window Proximity” users with Dark/Light/Contrasted lose that style on an empty desktop unless they enable the new option.

Suggestion: if (!GRADIENT_PANEL) keep PANEL_STYLES[STYLE_PANEL] like before; only apply gradient classes when the toggle is on.

src/styles/panel.css
```diff
@@ -35,6 +36,34 @@
...
+/* Internal use */
+
+#panel.gradient-panel-reverse {
+    background-color: transparent;
+    background-gradient-direction: vertical;
+    background-gradient-start: rgba(0,0,0,0);
+    background-gradient-end: rgba(0,0,0,0.5);
+    box-shadow: none !important;
+    transition-duration: 500ms;
+}
```
krissedout, line 65: Other panel styles use !important on background. These gradient rules don’t, so Shell themes that style #panel can override them and the gradient never shows.

Please add !important on background-color / gradient properties to match transparent-panel / dark-panel / light-panel.

src/extension.js
```diff
@@ -459,6 +459,20 @@ export default class BlurMyShell extends Extension {
...
+                this._panel_blur.reset();
+            }
+        });
+
+        this._settings.panel.GRADIENT_PANEL_MODE_changed(() => {
+            if (this._settings.panel.BLUR) {
+                this._panel_blur.connect_to_windows_and_overview();
+                this._panel_blur.reset();
+            }
+        });
```
krissedout, line 474: reset() disable+enable’s the panel blur on every gradient toggle/mode change. That’s heavier than STYLE_PANEL (connect only) and is a likely cause of the black flash you mentioned in the PR description.

For a style-class swap, update_visibility() / re-running set_should_override_panel should be enough without a full reset.

src/styles/panel.css
```diff
@@ -21,7 +22,7 @@
 * `.light-panel`
 */
 #panel.light-panel {
-    background-color: rgba(200, 200, 200, 0.2) !important;
+    background-color: rgba(200, 200, 200, 0.7) !important;
```
krissedout: This bumps light-panel from 0.2 → 0.7 globally (static override, mode 1, etc.), not only proximity + gradient.

If that was only meant for this feature, please scope it; if intentional everywhere, worth calling out explicitly in the PR.

resources/ui/panel.ui
```diff
@@ -202,6 +202,35 @@ Background on Window Proximity: Solid color background when a window is near the
...
+                  <object class="GtkSwitch" id="gradient_panel">
+                    <property name="valign">center</property>
+                  </object>
+                </child>
+              </object>
+            </child>
+
+            <child>
+              <object class="AdwActionRow" id="gradient_panel_mode_row">
+                <property name="title" translatable="yes">Panel position</property>
```
krissedout: Minor copy nits:
- “Panel position” reads like panel placement; “Gradient direction” matches the subtitle better.
- “Down” → “Bottom” to match the schema summary (0 Top, 1 Bottom).

resources/ui/panel.ui
```diff
@@ -276,6 +305,13 @@ Background on Window Proximity: Solid color background when a window is near the
     </items>
   </object>
 
+  <object class="GtkStringList" id="gradient_panel_mode_model">
+    <items>
+      <item translatable="yes">Top</item>
+      <item translatable="yes">Down</item>
```
krissedout: same thing as above comment

krissedout (changes requested):

MeguMario:
@krissedout Please check it again! Thank for the reviews!

krissedout (approved):
seems all good now! another thing we could do is add a panel text color override (light/dark) so people can do that instead of a gradient but that's not required for this obviously!

MeguMario:
@krissedout I've made some changes to fix the the few remaining issues (including the black flashing issues though it may still flash black if you move the window rapidly enough) so please check again if you can, thanks!

As for panel text color override, I've been thinking about it, though with how we have to consider that other extension can add their own icons as well, that's going to be difficult (lol).

src/extension.js
```diff
@@ -436,6 +436,7 @@ export default class BlurMyShell extends Extension {
         this._settings.panel.OVERRIDE_BACKGROUND_changed(() => {
             if (this._settings.panel.BLUR)
                 this._panel_blur.connect_to_windows_and_overview();
+                this._panel_blur.reset();
```
krissedout: Missing braces (again!) reset() always runs when OVERRIDE_BACKGROUND changes; only connect_to_windows_and_overview() is behind the BLUR check.

Please wrap both in { } like OVERRIDE_BACKGROUND_DYNAMICALLY_changed below. Also: master didn’t reset() on this handler, was that intentional, or leftover from the flash fix?

krissedout (changes requested):

src/extension.js
MeguMario: I always assume there are braces there for some reason so that's my oversight there (lol)

As for why `reset()`, it's to fix an issue where blur do not reapply when `OVERRIDE_BACKGROUND` is disabled with Blur on Window Proximity selected previously.

krissedout (approved):
alright all good now!

PR #932, opened 2026-07-19
Improve smoothness of certain Popup surface

MeguMario:
This improve smoothness of certain Popup surface by defer execution of signal connection 

https://github.com/user-attachments/assets/ed6e88a9-6efd-4ad5-9266-be73a027af21

This is still not as smooth as with popup blur off so there may be space to improve

### DIFF
--- a/src/components/popup/blur_surface.js
+++ b/src/components/popup/blur_surface.js
@@ -155,6 +155,11 @@ export const PopupBlurSurface = class PopupBlurSurface {
         const sibling_index = children.indexOf(sibling);
         return sibling_index >= 0 && actor_index < sibling_index;
     }
+    
+    is_excluded_from_deferring_surface() {
+        return this.style.has_any_style_class(this.target, ['datemenu-popover','modal-dialog'])
+            || this.style.has_any_style_class(this.root_actor, ['datemenu-popover','modal-dialog']);
+    }
 
     update() {
         try {
--- a/src/components/popup/surface_signals.js
+++ b/src/components/popup/surface_signals.js
@@ -1,4 +1,5 @@
 import * as Main from 'resource:///org/gnome/shell/ui/main.js';
+import GLib from 'gi://GLib';
 
 const SURFACE_SIGNALS = [
     'notify::allocation',
@@ -26,6 +27,7 @@ export const PopupBlurSurfaceSignals = class PopupBlurSurfaceSignals {
         this.signal_ids = [];
         this.signal_actors = new WeakSet();
         this.destroyed_actors = new WeakSet();
+        this.pendingIdles = new Set();
     }
 
     connect_actor(actor) {
@@ -35,10 +37,28 @@ export const PopupBlurSurfaceSignals = class PopupBlurSurfaceSignals {
         this.signal_actors.add(actor);
         this.track_destroy(actor);
 
+        const is_excluded_from_deferring_surface = this.surface.is_excluded_from_deferring_surface();
+
         SURFACE_SIGNALS.forEach(signal => {
             try {
-                const id = actor.connect(signal, () => this.surface.queue_update());
-                this.signal_ids.push([actor, id]);
+                let id = actor.connect(signal, () => {
+                    if (is_excluded_from_deferring_surface) {
+                        this.surface.queue_update();
+                    } else {
+                        if (this.pendingIdles.size > 0)
+                            return;
+
+                        let updateId = GLib.idle_add(GLib.PRIORITY_DEFAULT_IDLE, () => {
+                            this.pendingIdles.delete(updateId);
+                            this.surface.queue_update();
+                            return GLib.SOURCE_REMOVE;
+                        });
+
+                        if (updateId)
+                            this.pendingIdles.add(updateId);
+                    }
+                });
+                this.signal_ids.push([actor, id, signal]);
             } catch (e) { }
         });
     }
@@ -94,6 +114,14 @@ export const PopupBlurSurfaceSignals = class PopupBlurSurfaceSignals {
     }
 
     disconnect_all() {
+        if (this.pendingIdles.size > 0) {
+            for (const updateId of this.pendingIdles) {
+                if (updateId) {
+                    GLib.source_remove(updateId);
+                }
+            }
+            this.pendingIdles.clear();
+        };
         this.signal_ids.forEach(([signal_actor, signal_id]) => {
             if (this.destroyed_actors.has(signal_actor))
                 return;


src/components/popup/surface_signals.js
```diff
@@ -35,10 +51,15 @@ export const PopupBlurSurfaceSignals = class PopupBlurSurfaceSignals {
         this.signal_actors.add(actor);
         this.track_destroy(actor);
 
-        SURFACE_SIGNALS.forEach(signal => {
+        const is_quick_toggle = this.surface.is_quick_toggle_menu_surface();
+        const signals_to_connect = is_quick_toggle
+            ? ['notify::visible', 'notify::mapped', 'notify::opacity']
+            : SURFACE_SIGNALS;
```
krissedout: for quick-toggle this only connects visible / mapped / opacity, so nothing in GEOMETRY_SIGNALS is ever wired. That means suspend_geometry_signals / resume_geometry_signals are no-ops for this path, the win is the permanent signal cut, not the transition suspend, either connect the full (or geometry) set and suspend only while transition_state.running, or drop suspend/resume and keep this reduced set on purpose.

also: permanently dropping geometry + style-changed / pseudo-class means after the expand settles, blur only updates on visible/mapped/opacity. during the animation queue_transition_update still polls, but post-open layout changes without those notifies won’t refresh. worth checking open > settle > resize/content change. (note that it likely wont break EVERY resize but the post open layout changes can be missed)

src/components/popup/blur_surface.js
```diff
@@ -164,7 +170,18 @@ export const PopupBlurSurface = class PopupBlurSurface {
...
+            const is_quick_toggle_menu = this.is_quick_toggle_menu_surface();
             const visible = this.is_visible();
+
+            if (is_quick_toggle_menu && transition_state.running && !this._transitions_running) {
+                this._transitions_running = true;
+                try { this.signals.suspend_geometry_signals(); } catch (e) { }
+            }
+            if (this._transitions_running && !transition_state.running) {
+                this._transitions_running = false;
+                try { this.signals.resume_geometry_signals(); } catch (e) { }
```
krissedout: same issue from the call site: these suspend/resume calls never have geometry handlers to work with when connect_actor already skipped them (see surface_signals.js ~54–57). dead for the intended case.

src/components/popup/surface_signals.js
krissedout, line 133: nit: disconnect_all should also clear this.suspended_signals = [] so destroy can’t leave reconnect metadata around if suspend/resume is fixed later. (this isn't your change but still would be useful, even if it's harmless today)

krissedout (changes requested):

MeguMario:
@krissedout Thanks for the reviews! After some testing, I have decided to defer execution of signal connection instead of keeping the old method of suspending / cutting out signal since the old method break more than it help (suspending doesn't even work and cutting out signal is just like you said), this time it applys to most Popup surface with the exception of `datemenu-popover` since defering makes the animation looks kinda weird.

Please check it again if you have time!

src/components/popup/surface_signals.js
```diff
@@ -27,18 +28,35 @@ export const PopupBlurSurfaceSignals = class PopupBlurSurfaceSignals {
...
 
+        let updateId = 0;
+        const is_datemenu_surface = this.surface.is_datemenu_surface();
+
         SURFACE_SIGNALS.forEach(signal => {
             try {
-                const id = actor.connect(signal, () => this.surface.queue_update());
-                this.signal_ids.push([actor, id]);
+                if (is_datemenu_surface) {
+                    id = actor.connect(signal, () => this.surface.queue_update());
```
krissedout: id isn’t declared (let/const missing). In strict module code this throws after connect() succeeds; the catch swallows it so the signal never lands in signal_ids and disconnect_all can’t disconnect it. Use let id = actor.connect(...).

src/components/popup/surface_signals.js
```diff
@@ -27,18 +28,35 @@ export const PopupBlurSurfaceSignals = class PopupBlurSurfaceSignals {
...
+                }
+                else {
+                    id = actor.connect(signal, () => {
+                        if (updateId !== 0)
+                            return;
+
+                        updateId = GLib.idle_add(GLib.PRIORITY_DEFAULT_IDLE, () => {
```
krissedout: Please GLib.source_remove(updateId) from disconnect_all (and clear it) so a pending idle can’t outlive the surface.

src/components/popup/blur_surface.js
```diff
@@ -457,6 +463,10 @@ export const PopupBlurSurface = class PopupBlurSurface {
             this.corner_changed_id = 0;
         }
         this.fade = null;
+        if (this._transitions_running) {
```
krissedout: Leftover from the old suspend/resume path,  _transitions_running is never set and resume_geometry_signals is gone. Safe to delete this block.

krissedout (changes requested):

MeguMario:
@krissedout  I've made some modifications, please check again, thanks!

src/components/popup/surface_signals.js
```diff
@@ -35,10 +36,27 @@ export const PopupBlurSurfaceSignals = class PopupBlurSurfaceSignals {
...
+
         SURFACE_SIGNALS.forEach(signal => {
             try {
-                const id = actor.connect(signal, () => this.surface.queue_update());
-                this.signal_ids.push([actor, id]);
+                if (is_datemenu_surface) {
+                    let id = actor.connect(signal, () => this.surface.queue_update());
+                }
+                else {
+                    let id = actor.connect(signal, () => {
```
krissedout: Still broken, let id is scoped to the if/else blocks, so push([actor, id, signal]) throws and the empty catch hides it. Handlers connect but never enter signal_ids. Hoist: let id; before the if, assign in both branches, then push.

src/components/popup/surface_signals.js
```diff
@@ -94,6 +112,10 @@ export const PopupBlurSurfaceSignals = class PopupBlurSurfaceSignals {
     }
 
     disconnect_all() {
+        if (this.updateId !== 0) {
+            GLib.source_remove(this.updateId);
+            this.updateId = 0;
+        }
```
krissedout: Idle sources are stored in the local updateId from connect_actor, not this.updateId. This never cancels the real idle, and since this.updateId is unset, undefined !== 0 makes every disconnect call source_remove(undefined). Track pending idle(s) on this (Set if multiple actors) and guard with if (this.updateId).

krissedout (changes requested):

MeguMario:
@krissedout I've made some changes according to reviews, please review it again if you have time, thanks!

krissedout (approved):
lgtm! thanks for the fixes!

MeguMario:
@krissedout aight thanks a lot!

Small reminder to myself to not do these thing at 3am in the morning (lol).

PR #936, opened 2026-07-24
Use Meta.LaterType.IDLE to defer update for certain blur surface

MeguMario:
This PR add the use of ``Meta.LaterType.IDLE`` for dash-to-dock and popup-blur to defer blur surface update, replace the use of ``GLib.PRIORITY_DEFAULT_IDLE``.

### What has changed

- For popup blur, nothing changes.
- For dash-to-dock, this (kinda) fixed the issue of switching between dynamic and static blur result in blur breaking for the dock. Note that kinda means that it will still break, but it's much more usable than before, also one less import.

### DIFF
--- a/src/components/popup/surface_signals.js
+++ b/src/components/popup/surface_signals.js
@@ -1,5 +1,5 @@
+import Meta from 'gi://Meta';
 import * as Main from 'resource:///org/gnome/shell/ui/main.js';
-import GLib from 'gi://GLib';
 
 const SURFACE_SIGNALS = [
     'notify::allocation',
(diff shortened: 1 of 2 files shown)


krissedout (approved):
lgtm!

src/components/popup/surface_signals.js
```diff
@@ -45,17 +44,25 @@ export const PopupBlurSurfaceSignals = class PopupBlurSurfaceSignals {
...
-                        });
-
-                        if (updateId)
-                            this.pendingIdles.add(updateId);
+                        const laters = global.compositor?.get_laters?.();
+                        
+                        this.clear_pending_idles();
+                        
+                        if (laters) {
+                            // Modern GNOME Shell (GNOME 44+)
```
DaPigGuy, line 52: Post-merge FYI but we only support GNOME 46+ anyway, and supporting anything below 45 is absolutely not happening due to the move to ESM

MeguMario: Let me remove that real quick

MeguMario: Removed with https://github.com/aunetx/blur-my-shell/pull/939

PR #938, opened 2026-07-26
Fix static blur edge on bottom panels

Sultech:
## Summary

- extend the static panel blur clip upward by one pixel
- leave dynamic blur unchanged

## Problem

Static panel blur can leave a one-pixel unblurred strip along the workspace-facing edge of a bottom panel. Dynamic blur does not show the strip.

The static background is clipped exactly to the panel geometry. Expanding the clip upward by one pixel removes the rasterized edge without changing panel placement, size, or work-area geometry.

## Validation

- reproduced with Simple Taskbar on GNOME Shell 50.1
- confirmed that changing the background offset did not resolve the strip
- confirmed that expanding the static clip upward by one pixel removed it
- reverted the test to reproduce the original strip
- ran `make build`

Dash to Panel can show a similar, less visible edge, but the manual validation for this change was performed with Simple Taskbar.

### DIFF
--- a/src/components/panel.js
+++ b/src/components/panel.js
@@ -389,12 +389,19 @@ export const PanelBlur = class PanelBlur {
 
             let x = p_x + p_p_x - monitor.x + g_x;
             let y = p_y + p_p_y - monitor.y + g_y;
+            let is_horizontal = geometry_width >= geometry_height;
+            let distance_to_top = Math.abs(y);
+            let distance_to_bottom = Math.abs(
+                monitor.height - (y + geometry_height)
+            );
+            let is_bottom_panel = is_horizontal &&
+                distance_to_bottom < distance_to_top;
 
             background.set_clip(
                 x,
-                y - 1,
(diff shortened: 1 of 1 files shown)


krissedout (changes requested):
One issue: this applies to every static panel, not just bottom ones. On a top panel, `y - 1` / `height + 1` grows the clip the wrong way (and can spill a pixel into the workspace / off-monitor).

Can you gate this to the workspace-facing edge (or at least bottom panels), and re-check a normal top panel + Dash to Panel if easy?

Sultech:
> One issue: this applies to every static panel, not just bottom ones. On a top panel, `y - 1` / `height + 1` grows the clip the wrong way (and can spill a pixel into the workspace / off-monitor).
> 
> Can you gate this to the workspace-facing edge (or at least bottom panels), and re-check a normal top panel + Dash to Panel if easy?

I have gated it to the bottom panels, could you please take a look when you have time, thanks.

krissedout (approved):
seems good now

PR #941, opened 2026-07-27
Additional fixes for dash-to-dock blurring

### LINKED ISSUE #940: The static blur type only affects the dock on one monitor
When using the static blur type, only the dock on my main monitor gets the affect, the one on the secondary monitor just gets left with no background at all. But when using dynamic blur, both docks get the background just fine.

 \- What happens with static blur -
 
Image

v.s.

 \- What happens with dynamic blur -
 
Image

note - sorry for the long screenshots, I have a regular monitor with a wacom underneath.

MeguMario:
This fixes some additional bug for dash-to-dock blurring

1. This fully fix the issue where switching between blur mode on the dock can sometime fail
2. This fix the issue where static blur not blurring the dock in multi-monitor scenario (https://github.com/aunetx/blur-my-shell/issues/940) by instead calculating per dock position, now we will calculate the dock exact position to the entire multi-monitor canvas instead no matter the position.
3. This also fixed the issue where blur on dock can be seen on startup with Hide blur in Overview enabled
image

### DIFF
--- a/src/components/dash_to_dock.js
+++ b/src/components/dash_to_dock.js
@@ -36,6 +36,14 @@ class DashInfos {
         this.old_style = this.dash._background?.style;
 
         this.updateId = 0;
+        this._first_boot = true;
+
+        this.bg_allocation_id = this.background_group?.connect('notify::allocation', () => {
+            this.schedule_update();
+        });
+
+        let monitor = Main.layoutManager.findMonitorForActor(this.dash_container);
+        this.current_monitor_index = monitor ? monitor.index : null;
 
         this.dash_destroy_id = dash.connect('destroy', () => this.remove_dash_blur(false));
         this.dash_blur_connections_ids = [];
@@ -119,6 +127,11 @@ class DashInfos {
             this.bg_manager._bms_pipeline.destroy();
             this.bg_manager._bms_pipeline = null;
         }
+        if (this.bg_allocation_id && this.background_group) {
+            this.background_group.disconnect(this.bg_allocation_id);
+            this.bg_allocation_id = null;
+        }
+        
         this.bg_manager?.destroy();
         this.background_group?.destroy();
     }
@@ -126,7 +139,7 @@ class DashInfos {
     change_blur_type() {
         this.destroy_dash();
 
-        let blur_result = this.dash_blur.add_blur(this.dash);
+        let blur_result = this.dash_blur.add_blur(this.dash, this.dash_container);
         if (!blur_result)
             return;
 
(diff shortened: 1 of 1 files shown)


src/components/dash_to_dock.js
```diff
@@ -52,11 +55,33 @@ class DashInfos {
     }
 
     schedule_update() {
-        this.clear_pending_idles();
+        if (this.updateId)
+            return;
+
+        let frames_still = 0;
+        let last_x = this.dash_container.x;
```
krissedout: This tracks the outer container, but Dash-to-Dock animates _slider.slide-x and relayouts its child while dash_container.x/y can remain unchanged. We can therefore declare the dock stable after three idle cycles while it is still animating. Could this track the slider’s allocation/transformed position or slide-x instead?

src/components/dash_to_dock.js
```diff
@@ -208,26 +242,26 @@ class DashInfos {
...
-            y = dash_container.y + (dash_container.height - dash_background.height) / 2 - dash_background.y;
-        }
+        let parent = this.background_group?.get_parent();
+            if (!parent) return;
+
+        let [parent_stage_x, parent_stage_y] = parent.get_transformed_position();
+        let [bg_stage_x, bg_stage_y] = this.dash_background.get_transformed_position();
 
-        return [x, y];
+        this.background.x = monitor.x - parent_stage_x;
```
krissedout, line 232: This correctly computes the full-monitor background origin relative to its parent, but update_size() immediately overwrites it with this.background.x = -x / y = -y, where x/y are clip coordinates. The side branches then add dash offsets to coordinates already derived from the transformed background. Could get_dash_position() return {backgroundX, backgroundY, clipX, clipY} and have update_size() apply each exactly once?

krissedout (changes requested):

src/components/dash_to_dock.js
MeguMario: it's now apply exactly once

src/components/dash_to_dock.js
MeguMario: I changed my approach for this so please check again

src/components/dash_to_dock.js
```diff
@@ -36,6 +36,14 @@ class DashInfos {
         this.old_style = this.dash._background?.style;
 
         this.updateId = 0;
+        this._first_boot = true;
+
+        this.bg_allocation_id = this.background_group?.connect('notify::allocation', () => {
+            this.update_size();
```
krissedout: This calls update_size() synchronously from notify::allocation, while update_size() changes actor position and clipping. We deliberately defer these geometry mutations through schedule_update()/Meta.LaterType.IDLE; the same immediate-update pattern previously caused “Can’t update stage views” allocation spam in the panel path. Could both background-group allocation callbacks use this.schedule_update() instead? The duplicate callback in change_blur_type() has the same issue.

krissedout (changes requested):

src/components/dash_to_dock.js
MeguMario: both now use `this.schedule_update()`, forgot to replace those after testing.

krissedout (approved):
LGTM now!

PR #920, opened 2026-07-05
feat(shaders): GNOME 51 Cogl.Snippet support

krissedout:
GNOME 51 removes `CoglProgram`/`CoglShader` and ports `Clutter.ShaderEffect` to `Cogl.Snippet`, dropping `set_shader_source` in favor of `get_static_snippet()`:

- [mutter !5094 — Get rid of CoglProgram/CoglShader](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/5094)
- [gnome-shell !4241 — Use ClutterShaderEffect instead of GLSLEffect](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/4241)

Without this change, the extension fails to load on Shell 51 with `get_static_snippet` errors.

(For clarity: [mutter !5071](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/5071) `ext-background-effect-v1` is a separate compositor protocol for client surfaces and cannot be used here because BMS runs inside Shell and Shell performs special blurring.)

## Summary

- Adds runtime dual-path shader support for GNOME 51 (`Cogl.Snippet` + `vfunc_get_static_snippet`) while preserving the legacy `set_shader_source` path for Shell 46–50
- Centralizes shader registration and only installs the snippet vfunc when the runtime supports it, so GNOME 46–50 never see an unknown virtual function
- Fixes a 1px unblurred seam on static panel blur with snippet offscreen buffers (integer positioning + clip inset on 51 only)
- Cleans up pooled-effect actor lifecycle handling and releases managed effects when the extension is disabled
- Detects native `corner-radius` support from the active blur implementation, including the compatibility typelib and future Shell implementations

## Technical notes

- API detection uses `Clutter.ShaderEffect.new_with_snippet` at runtime, not `metadata.json` or `PACKAGE_VERSION`
- GNOME 50 panel behavior is unchanged (`subpixel_stage_offset = 0.5`, `static_blur_clip_inset = 0`)
- Dash-to-dock keeps its original vertical alignment on 50; only receives clip inset on 51
- Dynamic rounded blur uses the native blur property when available; static rounded blur remains built into BMS

PR #944, opened 2026-08-01
Some optimizations to Quick Settings blurring

MeguMario:
**NOTE:** This is tested with optimization found in https://github.com/aunetx/blur-my-shell/pull/920

This PR introduces a few optimization to mostly further speedup Quick Settings blurring (though some of them can be apply to general popup blur as well)

- Cut down the amount of signal connect to Quick settings and `datemenu-popover` specifically (again)
- Fixed the issue where `quick-toggle-menu` can last for some time after it was dismissed

Originally this was going to come with an update throttler and interpolation via Clutter's `AnimationMode` but I feel like it introduce way too much complexity and didn't actually fix the problem.

Now does it actually speedup quick settings, not sure, I tested it on my computer and it seems to be quite smooth on battery, but on older hardware, idk

### DIFF
--- a/src/components/popup/blur_surface.js
+++ b/src/components/popup/blur_surface.js
@@ -155,9 +155,15 @@ export const PopupBlurSurface = class PopupBlurSurface {
         return sibling_index >= 0 && actor_index < sibling_index;
     }
     
-    is_excluded_from_deferring_surface() {
-        return this.style.has_any_style_class(this.target, ['datemenu-popover','modal-dialog','quick-settings'])
-            || this.style.has_any_style_class(this.root_actor, ['datemenu-popover','modal-dialog','quick-settings']);
+
+    is_quick_settings() {
+        return this.style.has_any_style_class(this.target, ['quick-toggle-menu','quick-settings'])
+            || this.style.has_any_style_class(this.root_actor, ['quick-toggle-menu','quick-settings']);
+    }
+
+    is_heavy_surface() {
+        return this.style.has_any_style_class(this.target, ['datemenu-popover','quick-settings','modal-dialog'])
+            || this.style.has_any_style_class(this.root_actor, ['datemenu-popover','quick-settings','modal-dialog']);
     }
 
     update() {
--- a/src/components/popup/surface_signals.js
+++ b/src/components/popup/surface_signals.js
@@ -21,6 +21,24 @@ const SURFACE_SIGNALS = [
     'style-changed',
 ];
 
+const ANCESTOR_SIGNALS = [
+    'notify::allocation',
+    'notify::position',
+    'notify::size',
+    'notify::x',
+    'notify::y',
+    'notify::width',
+    'notify::height',
+    'notify::clip-rect',
+    'notify::visible',
+    'notify::mapped',
+    'notify::translation-x',
+    'notify::translation-y',
+    'notify::scale-x',
+    'notify::scale-y',
+    'notify::pseudo-class',
+];
+
 export const PopupBlurSurfaceSignals = class PopupBlurSurfaceSignals {
     constructor(surface) {
         this.surface = surface;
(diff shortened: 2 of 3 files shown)


src/components/popup/surface_signals.js
```diff
@@ -21,6 +21,13 @@ const SURFACE_SIGNALS = [
     'style-changed',
 ];
 
+const ANCESTOR_SIGNALS = [
```
krissedout, line 24: Ancestor transforms still affect the target’s stage-space geometry. Restricting ancestors to allocation, clip, and visibility drops translation and scale changes, which are not guaranteed to emit notify::allocation. A popup whose parent animates can therefore leave its blur at stale coordinates. Please retain the geometry and transform signals for ancestors; opacity and style signals can still be omitted.

src/components/popup/index.js
```diff
@@ -304,8 +304,7 @@ export const PopupBlur = class PopupBlur {
         try {
             const class_names = actor.get_style_class_name?.();
             if (typeof class_names === 'string') {
-                const normalized = ` ${class_names.trim().replace(/\s+/g, ' ')} `;
-                return style_classes.some(style_class => normalized.includes(` ${style_class} `));
+                return style_classes.some(style => class_names.includes(style));
```
krissedout: Style classes are whitespace-delimited tokens, not substrings. This makes a class such as bms-popup-blurred-widget-extra match bms-popup-blurred-widget, causing legitimate actors to be treated as internal and skipped. Please keep exact token matching with split(/\s+/).includes(...) or a Set.

src/components/popup/blur_surface.js
```diff
@@ -155,9 +155,9 @@ export const PopupBlurSurface = class PopupBlurSurface {
         return sibling_index >= 0 && actor_index < sibling_index;
     }
     
-    is_excluded_from_deferring_surface() {
-        return this.style.has_any_style_class(this.target, ['datemenu-popover','modal-dialog','quick-settings'])
-            || this.style.has_any_style_class(this.root_actor, ['datemenu-popover','modal-dialog','quick-settings']);
+    is_heavy_surface() {
+        return this.style.has_any_style_class(this.target, ['datemenu-popover','quick-settings'])
```
krissedout: Why was modal-dialog removed from the non-deferred surfaces? Its geometry changes now take the IDLE path and then schedule another before-redraw update, potentially putting the backdrop one frame behind dialog animations. Please preserve the existing modal-dialog behavior unless this has been runtime-tested.

src/components/popup/surface_placement.js
```diff
@@ -339,5 +339,8 @@ export const PopupBlurSurfacePlacement = class PopupBlurSurfacePlacement {
         this.surface_height = null;
         this.monitor_index = null;
         this.offscreen = false;
+
+        this._cached_parent_x = undefined;
```
krissedout: _cached_parent_x and _cached_parent_y are not referenced anywhere else, so these assignments have no effect and should be removed.

krissedout (changes requested):

MeguMario:
@krissedout A little comment here to say that I cannot make any update on this for now due to my computer is currently not functioning right now, will return to this ASAP once I get that sort out

Sorry for the trouble.

krissedout:
@MeguMario it's ok, dont worry about it there's no rush!

src/components/popup/surface_signals.js
MeguMario: It's now only omitted opacity and style signals

src/components/popup/surface_placement.js
MeguMario: I've cleaned these up, forgot about those.

src/components/popup/index.js
MeguMario: Reverted to the previous implementation

src/components/popup/blur_surface.js
MeguMario: Reverted now

Hazem-Samir:
I want to ask unrelated question to @MeguMario , when you gonna add the glass effect to dynamic blur?
 I am really exciting to see it

src/components/popup/surface_signals.js
```diff
@@ -29,30 +47,32 @@ export const PopupBlurSurfaceSignals = class PopupBlurSurfaceSignals {
         this.destroyed_actors = new WeakSet();
     }
 
-    connect_actor(actor) {
+    connect_actor(actor, is_ancestor = false) {
```
krissedout, line 50: is_ancestor / ANCESTOR_SIGNALS are added here, but connect_ancestors still calls connect_actor(actor) with the default false, so ancestors never use ANCESTOR_SIGNALS. Please also update that call site to this.connect_actor(actor, true).

src/components/popup/surface_signals.js
```diff
@@ -29,30 +47,32 @@ export const PopupBlurSurfaceSignals = class PopupBlurSurfaceSignals {
...
+                const is_visibility_change = signal === 'notify::visible' || signal === 'notify::mapped';
+                const laterType = (is_heavy_surface || is_visibility_change) 
+                    ? Meta.LaterType.BEFORE_REDRAW
+                    : Meta.LaterType.IDLE;
+
+                this.updateId = global.compositor.get_laters().add(laterType, () => {
+                    this.updateId = 0;
+                    this.surface.queue_update();
+                    return false;
                 });
```
krissedout: Heavy/visibility paths now schedule a later that then calls queue_update(), and queue_update() schedules another BEFORE_REDRAW. That adds an extra hop vs the old direct this.surface.queue_update() fast path for heavy surfaces. Please call queue_update() directly for heavy/visibility, and keep the IDLE coalesce only for light surfaces.

src/components/popup/surface_transitions.js
```diff
@@ -79,8 +79,10 @@ export const PopupBlurSurfaceTransitions = class PopupBlurSurfaceTransitions {
         const seen = new WeakSet();
         actors.forEach(actor => seen.add(actor));
 
-        this.add_descendants(actors, seen, this.surface.target);
-        this.add_descendants(actors, seen, this.surface.root_actor);
+        if (!this.surface.is_heavy_surface()) {
+            this.add_descendants(actors, seen, this.surface.target);
+            this.add_descendants(actors, seen, this.surface.root_actor);
+        }
```
krissedout: Skipping descendant transition scans fixes lingering quick-toggle-menu blur, but is_heavy_surface() also covers modal-dialog and datemenu-popover. Nested fades on those can hide the blur early. Consider limiting this skip to quick-settings / quick-toggle-menu.

krissedout (changes requested):

src/components/popup/surface_transitions.js
MeguMario: limited these to both `quick-settings` and `quick-toggle-menu`

src/components/popup/surface_signals.js
MeguMario: `is_heavy_surface` now call `queue_update()` directly

src/components/popup/surface_signals.js
MeguMario: forgot about that one, added!

krissedout (approved):
lgtm now from the code side

MeguMario:
> lgtm now from the code side

I tested it on my (now dead) computer and the current one (which is weaker than the last one) and it seems much smoother than before, there doesn't seems to be any bug both with stock settings or with other extensions so I merged it.

PR #946, opened 2026-08-01
refactor: streamline runtime lifecycle and settings

krissedout:
## Summary

- make pipeline settings parsing and serialization atomic
- prevent duplicated pipelines from sharing or mutating effect data
- correctly synchronize externally created, renamed, and removed pipelines
- fix pipeline actor signal cleanup and redundant effect removal
- group managed connections by object instead of installing a destroy handler per signal
- remove the per-frame GObject signal round-trip from dynamic blur repaint handling
- reuse app-folder blur effects and replace legacy Tweener animation with `Clutter.Timeline`
- use constant-time effect pool operations
- fix unnecessary pipeline selector list rebuilding
- remove dead code and split pipeline serialization and effect-group metadata into focused modules

## Why

Several runtime helpers had accumulated avoidable allocations, duplicate signal handlers, and validation paths that could continue after encountering malformed data. Pipeline duplication could also mutate the original pipeline because effect objects were shared.

This refactor tightens those lifecycle boundaries while preserving the existing dynamic blur repaint workaround and GNOME 46–51 compatibility.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ install: build remove
 
 pot:
 	find resources/ui -iname "*.ui" -printf "%p\n" | sort | \
-		xargs xgettext --output=po/$(UUID).pot src/effects/effects.js \
+		xargs xgettext --output=po/$(UUID).pot src/effects/effects.js src/effects/effect_groups.js \
 		--from-code=utf-8 --package-name=$(UUID)
 
 	rm po/LINGUAS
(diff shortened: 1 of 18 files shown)


MeguMario (approved):
lgtm. I also ran the change on my end and seems like there are no issue

I'll leave the merging to you just in case you still have something in mind.

PR #947, opened 2026-08-02
implemented notif cards clipping masks for stacked notifications

### LINKED ISSUE #943: Feature request: Blur for notification cards
Now that popup blur has shipped, I'd love to see notification cards get the same treatment next. If someone's using a shell theme with transparent notifications, unblurred cards become a mess once more than one is stacked — you end up seeing the transparent notifications underneath the top one, with their text and outlines all jumbled together. Blurring notification cards would keep stacks legible the same way popup blur already does for popups.

P.S also make the popup blur persistent over in the lockscreen, popups aren't blurred there

rinzler69-wastaken:
a PR made to close issue #943 by me if approved

for when Popup Blur is ON, and notif cards are translucent, i implemented clipping mask so that the notif cards beneath the top one does not show through

### DIFF
--- a/src/components/popup/message_stacks.js
+++ b/src/components/popup/message_stacks.js
@@ -14,6 +14,8 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
         this.queued_actors = new WeakSet();
         this.watched_actors = new WeakSet();
         this.destroyed_actors = new WeakSet();
+        this.bms_clipped = new WeakSet();
+        this.group_connections = new Map();
         this.update_ids = new Map();
         this.original_opacity = new WeakMap();
         this.enabled = false;
@@ -136,12 +138,23 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
         );
 
         if (group.layout_manager) {
-            this.connect(
-                group.layout_manager,
-                'notify::expansion',
-                () => this.update_group_header(group)
-            );
+            const lm = group.layout_manager;
+            try {
+                const id = lm.connect('notify::expansion', () => this.update_group_header(group));
+                this.group_connections.set(group, { lm, id });
+            } catch (e) { }
         }
+    }
+
+    untrack_group(group) {
+        if (this.group_connections.has(group)) {
+            const { lm, id } = this.group_connections.get(group);
+            try {
+                lm.disconnect(id);
+            } catch (e) { }
+            this.group_connections.delete(group);
+        }
+        this.groups.delete(group);
     }
 
     update_all() {
@@ -162,11 +175,6 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
             const target_opacity = this.enabled ? Math.round(expansion * 255) : 255;
 
             header.opacity = target_opacity;
-
-            header.get_children?.().forEach(actor => {
-                if (actor)
-                    actor.opacity = target_opacity;
-            });
         } catch (e) { }
     }
 
@@ -204,7 +212,6 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
             this.hide_message_content(message);
         } else {
             this.remove_stack_mask(message);
-            this.restore_message_content(message);
         }
     }
 
@@ -226,20 +233,24 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
                 const visible_edge = 10;
                 const clip_y = Math.max(0, height - visible_edge);
                 message.set_clip(0, clip_y, width, visible_edge);
+                this.bms_clipped.add(message);
             } else if (index === 2) {
                 // 3rd card: tight visible strip of 7px (matches exact 7px bottom offset below card 2)
                 const visible_edge = 7;
                 const clip_y = Math.max(0, height - visible_edge);
                 message.set_clip(0, clip_y, width, visible_edge);
+                this.bms_clipped.add(message);
             } else if (index >= 3) {
                 // 4th+ cards: hidden when collapsed so they don't stack behind card 3
                 message.set_clip(0, height, width, 0);
+                this.bms_clipped.add(message);
             } else {
                 // Fallback by pseudo class
                 const is_second = this.has_pseudo_class(message, 'second-in-stack');
                 const visible_edge = is_second ? 10 : 6;
                 const clip_y = Math.max(0, height - visible_edge);
                 message.set_clip(0, clip_y, width, visible_edge);
+                this.bms_clipped.add(message);
             }
         } catch (e) { }
     }
@@ -248,13 +259,12 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
         if (!message || this.destroyed_actors.has(message))
             return;
 
-        try {
-            message.remove_clip();
-        } catch (e) { }
-
-        try {
-            message.get_parent?.()?.remove_clip();
-        } catch (e) { }
+        if (this.bms_clipped.has(message)) {
+            try {
+                message.remove_clip();
+            } catch (e) { }
+            this.bms_clipped.delete(message);
+        }
 
         this.restore_message_content(message);
     }
@@ -281,8 +291,6 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
             if (this.original_opacity.has(child)) {
                 child.opacity = this.original_opacity.get(child);
                 this.original_opacity.delete(child);
-            } else {
-                child.opacity = 255;
             }
         } catch (e) { }
     }
@@ -427,7 +435,9 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
             this.connections.connect(actor, 'destroy', () => {
                 this.destroyed_actors.add(actor);
                 this.containers.delete(actor);
-                this.groups.delete(actor);
+                if (this.groups.has(actor)) {
+                    this.untrack_group(actor);
+                }
                 this.messages.delete(actor);
                 this.cancel_update(actor);
             });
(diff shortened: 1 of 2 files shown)


rinzler69-wastaken:
image

rinzler69-wastaken:
the last commit i made to my branch is to solve these issues:

1. Visual Refinements to Notif Stack: when popup blur is ON, and background is overridden, the notif cards would go transparent. and as you collapse it, you will still see the notif stack app name getting trampled by the transparent notif cards then blip out. i made the app name and collapse button fade in and out as you expand or collapse the notif cards. this is only active when popup blur is ON

2. Visual Refinements to #notification-banner notif cards blur: when a notification card slides down, as the blur layer slides along with the card, it steps on the clock, blurring it for a splitsecond. i made the blur layer fade in and out as they slide down and back up.

src/components/popup/message_stacks.js
```diff
@@ -147,22 +191,148 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
...
+        const is_expanded = group ? group.expanded : this.is_in_expanded_group(message);
+
+        if (this.is_stacked(message) && !is_expanded) {
+            this.apply_stack_mask(message);
+            this.hide_message_content(message);
+        } else {
+            this.remove_stack_mask(message);
+            this.restore_message_content(message);
+        }
+    }
```
krissedout, line 215: This introduces persistent clipping but the component’s disable() path still only calls restore_message(), which restores opacity without removing the clip. Disabling popup blur while notifications are collapsed therefore leaves those messages clipped. Please call remove_stack_mask() for every tracked message before clearing the state.

src/components/popup/message_stacks.js
```diff
@@ -147,22 +191,148 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
...
+        const group = this.get_message_group(message);
+        const is_expanded = group ? group.expanded : this.is_in_expanded_group(message);
+
+        if (this.is_stacked(message) && !is_expanded) {
+            this.apply_stack_mask(message);
+            this.hide_message_content(message);
+        } else {
+            this.remove_stack_mask(message);
+            this.restore_message_content(message);
+        }
```
krissedout, line 214: remove_stack_mask() already calls restore_message_content(). Calling it again here finds that the saved opacity has been deleted and falls through to forcing 255, so the original opacity is always discarded. Please restore the content only once and never force an opacity that BMS did not save.

src/components/popup/message_stacks.js
```diff
@@ -147,22 +191,148 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
...
+        if (!message || this.destroyed_actors.has(message))
+            return;
+
+        try {
+            message.remove_clip();
+        } catch (e) { }
+
+        try {
+            message.get_parent?.()?.remove_clip();
+        } catch (e) { }
```
krissedout: This code removes the parent’s clip even though this PR never installs one there. GNOME Shell or the active theme may own that clip for layout or animation, so expanding a notification can erase unrelated clipping state. Please only remove clipping installed by BMS, and preserve any pre-existing clip on the message before replacing it.

src/components/popup/message_stacks.js
```diff
@@ -114,16 +121,53 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
...
+            () => this.update_group_header(group)
+        );
+
+        if (group.layout_manager) {
+            this.connect(
+                group.layout_manager,
+                'notify::expansion',
+                () => this.update_group_header(group)
+            );
+        }
```
krissedout, line 155: This connection is owned by the layout manager, but destroying the notification group only removes the group from the tracking set. The handler therefore retains the destroyed group until popup blur is disabled, accumulating connections as groups are recreated. Please retain the connection ID and disconnect it when the group is destroyed.

src/components/popup/message_stacks.js
```diff
@@ -114,16 +121,53 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
...
+
+            const expansion = group.layout_manager?.expansion ?? (group.expanded ? 1 : 0);
+            const target_opacity = this.enabled ? Math.round(expansion * 255) : 255;
+
+            header.opacity = target_opacity;
+
+            header.get_children?.().forEach(actor => {
+                if (actor)
+                    actor.opacity = target_opacity;
+            });
```
krissedout: Setting the same opacity on both the header and all its children multiplies the fade twice, making the effective fade quadratic. It also leaves child opacity modified after popup blur is disabled. Setting the header opacity alone already fades its descendants and avoids mutating their state.

src/components/popup/message_stacks.js
```diff
@@ -330,6 +500,14 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
...
         this.update_ids.clear();
         this.messages.forEach(message => this.restore_message(message));
+        this.groups.forEach(group => {
+            try {
+                if (group._headerBox) {
+                    group._headerBox.remove_all_transitions?.();
+                    group._headerBox.opacity = 255;
+                }
+            } catch (e) { }
+        });
```
krissedout, line 487: BMS does not create any transitions on _headerBox, so remove_all_transitions() can cancel animations owned by GNOME Shell or the theme. This cleanup also restores only the header while the child opacities modified above remain unchanged. Please track and restore only state introduced by BMS.

src/components/popup/surface_fade.js
```diff
@@ -32,9 +32,44 @@ export const PopupBlurSurfaceFade = class PopupBlurSurfaceFade {
...
+            const [stage_x, stage_y] = this.target.get_transformed_position();
+            const panel_height = 36;
+            const threshold = panel_height + 5;
+
+            if (stage_y <= 0)
+                return 0;
+            if (stage_y >= threshold)
+                return 1;
+
+            return Math.clamp(stage_y / threshold, 0, 1);
```
krissedout: The panel is not guaranteed to be 36 pixels tall; themes and panel extensions can change its geometry. This can fade the blur too early or still allow it to overlap the clock. Please calculate the threshold from the panel’s transformed bottom edge instead of using a fixed value.

krissedout (changes requested):

rinzler69-wastaken:
addressed all 7 code review comments just now, hope it solves your concerns

src/components/popup/surface_fade.js
```diff
@@ -32,9 +34,51 @@ export const PopupBlurSurfaceFade = class PopupBlurSurfaceFade {
...
+                panel_bottom = panel_y + panel_height;
+            }
+
+            const threshold = panel_bottom + 5;
+
+            if (stage_y <= 0)
+                return 0;
+            if (stage_y >= threshold)
+                return 1;
+
```
krissedout: This calculation now handles different panel heights, but still assumes the panel is at the top. With a bottom panel, panel_bottom is near the screen bottom, so banners near the top receive an alpha close to zero. Please calculate actual overlap with a top-edge panel rather than dividing absolute stage_y by panel_bottom.

krissedout (changes requested):

src/components/popup/surface_fade.js
rinzler69-wastaken: just updated ```get_banner_position_alpha()``` to first see that ```Main.panel``` is anchored near the top of the screen ```(panel_y <= 50)``` so setups with bottom panels return full alpha without fading

src/components/popup/surface_fade.js
```diff
@@ -1,3 +1,5 @@
+import * as Main from 'resource:///org/gnome-shell/ui/main.js';
```
krissedout: This resource URI is invalid. GNOME Shell exposes main.js at resource:///org/gnome/shell/ui/main.js, not resource:///org/gnome-shell/ui/main.js. This was fixed in the previous commit but has been reintroduced here, and it prevents the extension module from loading.

src/components/popup/surface_fade.js
```diff
@@ -32,9 +34,55 @@ export const PopupBlurSurfaceFade = class PopupBlurSurfaceFade {
...
+
+            const panel_bottom = panel_y + panel_height;
+            const threshold = panel_bottom + 5;
+
+            if (stage_y <= panel_y)
+                return 0;
+            if (stage_y >= threshold)
+                return 1;
+
+            return Math.clamp((stage_y - panel_y) / (threshold - panel_y), 0, 1);
```
krissedout, line 90: Checking only panel_y misclassifies left/right vertical panels as top panels because they also begin near y = 0. Their height is approximately the monitor height, so the notification blur remains almost entirely transparent. Global panel_y > 50 also fails when the primary monitor is positioned below another monitor. Please compare the panel’s transformed bounds with its monitor’s bounds and only apply this fade when it is a horizontal panel on that monitor’s top edge.

krissedout (changes requested):

src/components/popup/message_stacks.js
```diff
@@ -329,12 +509,30 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
     disable() {
         this.update_ids.forEach(id => GLib.source_remove(id));
         this.update_ids.clear();
-        this.messages.forEach(message => this.restore_message(message));
+        this.messages.forEach(message => {
+            this.remove_stack_mask(message);
+            this.restore_message(message);
+        });
```
krissedout: remove_stack_mask() now restores the saved child opacity, so the immediately following restore_message() is redundant. The same pairing exists in the destroy handler. Can we remove the redundant calls and the now-unused restore_message() / restore_child() helpers?

src/components/popup/surface_fade.js
```diff
@@ -32,9 +34,65 @@ export const PopupBlurSurfaceFade = class PopupBlurSurfaceFade {
...
+            if (!Main.panel)
+                return 1;
+
+            const monitor = Main.layoutManager.findMonitorForActor(Main.panel)
+                ?? Main.layoutManager.primaryMonitor;
+
+            if (!monitor)
+                return 1;
+
+            const [panel_x, panel_y] = Main.panel.get_transformed_position();
```
krissedout: nit: stage_x and panel_x are unused. Please destructure only the required coordinates: const [, stage_y] = … and const [, panel_y] = ….

src/components/popup/surface_fade.js
rinzler69-wastaken: I've updated get_banner_position_alpha() to retrieve Main.panel's monitor via findMonitorForActor() and verify that the panel is both horizontal (panel_width > panel_height) and anchored at its monitor's top edge (Math.abs(panel_y - monitor.y) <= 5). Vertical side panels and non-top panels now return full alpha immediately, while top-edge panels calculate relative overlap against panel_y and panel_bottom correctly across multi-monitor setups

krissedout (changes requested):

src/components/popup/message_stacks.js
krissedout: get_opacity() and set_opacity() no longer have any callers after restore_child() was removed. Please remove these dead helpers as part of the cleanup. After this, I think we can merge!

krissedout (changes requested):
Alright LGTM now, just one more thing.

src/components/popup/message_stacks.js
rinzler69-wastaken: Removed the dead get_opacity() and set_opacity() helpers from message_stacks.js... Thanks for all the thorough reviews 🙌

krissedout (approved):

PR #950, opened 2026-08-03
Complete Arabic translation and fix minor issues

HamzaSyrage:
This PR completes the Arabic translation of the extension and fixes a few minor issues in the existing localization.

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -111,7 +111,7 @@ msgstr "التكرارات"
 
 #: src/effects/effects.js:133
 msgid "The number of iterations. The more there are, the smoother the blur is."
-msgstr "عدد التكرارات. كلما زاد، كلما كان التموية املس."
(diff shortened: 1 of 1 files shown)


krissedout:
Thanks for this! Weblate is currently stuck/locked because of a repo sync conflict, and we use Hosted Weblate for translations. Could you hold off / redo these Arabic edits there once it's unlocked? That way they won't conflict with the Weblate sync again. I'll ping here when it's good to go. Also you seem to have a copilot suggestion :)

HamzaSyrage:
ok sure  Just ping me once Weblate is ready.
also about the Copilot suggestion Arabic script works differently from Latin script so having concatenated strings without adding whitespace isn't actually an issue in this case. will fix it anyway and I also turned off the automatic Copilot code reviews :)

aunetx:
Thanks @HamzaSyrage! I just unlocked weblate, so you may want to do that from there, thanks for your work :)

MeguMario:
With weblate unlocked I guess we can close this now

PR #955, opened 2026-08-06
refine notifcard expand/collapse look and feel for Popup Blur mode ON

rinzler69-wastaken:
## Description
This PR resolves visual artifacts and harsh color flashes when expanding or collapsing notification card stacks with translucent background styling and especially when Popup Blur is ON.

### Problem
Previously, when expanding a collapsed notification group, `remove_stack_mask()` was invoked immediately at the start of the expansion transition. This unclipped all 10–20 notification cards and restored their content opacities to 100% while they were still stacked on top of each other at the top of the list. As a result, 10–20 translucent card background layers accumulated simultaneously, producing a bright, solid color flash during the unraveling animation. Similarly, during collapse, cards were visibly masked while sliding upward.

### Solution
- **Expansion-Aware Tracking**: Connected `notify::expansion` on the layout manager to track `expansion` frame-by-frame ($0.0 \to 1.0$) during the transition animation.
- **Dynamic Clip & Opacity Interpolation**:
  - `0.01 < expansion < 0.99`: Clip heights expand dynamically in sync with the layout animation, while card content and background opacities are staggered and faded in proportionally to `expansion`.
  - `expansion >= 0.99`: Removes clips cleanly and enforces 100% opacities once fully unravelled.
  - `expansion <= 0.01`: Restores the tight collapsed mask heights and hidden child opacities.

### Benefits
- **Zero Alpha Pile-up**: Prevents translucent background accumulation during group unraveling.
- **Smooth Collapsing**: Cards shrink clips and fade out gracefully when bundling back up, eliminating "masked cards flying upward" glitches.
- **High Performance**: Performs basic math in JS without triggering layout reflows or CSS invalidations; reduces GPU overdraw during transitions.

### Before:
Screenshot From 2026-08-06 22-36-10
Screenshot From 2026-08-06 22-12-06

### After (Behavior of both Expand and Collapse):
Screenshot From 2026-08-06 22-38-55

### DIFF
--- a/src/components/popup/message_stacks.js
+++ b/src/components/popup/message_stacks.js
@@ -411,8 +411,6 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
             if (this.original_message_opacity.has(message)) {
                 message.opacity = this.original_message_opacity.get(message);
                 this.original_message_opacity.delete(message);
-            } else {
-                message.opacity = 255;
             }
         } catch (e) { }
     }
@@ -616,6 +614,7 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
         this.destroyed_actors = new WeakSet();
         this.bms_clipped = new WeakSet();
         this.original_opacity = new WeakMap();
+        this.original_message_opacity = new WeakMap();
         this.enabled = false;
     }
 };
--- a/src/components/popup/message_stacks.js
+++ b/src/components/popup/message_stacks.js
@@ -18,6 +18,7 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
         this.group_connections = new Map();
         this.update_ids = new Map();
         this.original_opacity = new WeakMap();
+        this.original_message_opacity = new WeakMap();
         this.enabled = false;
     }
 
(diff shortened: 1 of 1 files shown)


src/components/popup/message_stacks.js
```diff
@@ -294,6 +393,30 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
...
+    restore_message_opacity(message) {
+        if (!message || this.destroyed_actors.has(message))
+            return;
+        try {
+            if (this.original_message_opacity.has(message)) {
+                message.opacity = this.original_message_opacity.get(message);
+                this.original_message_opacity.delete(message);
+            } else {
+                message.opacity = 255;
+            }
```
krissedout: When no original was stored, restore_message_opacity forces message.opacity = 255. That runs for index < 3 on every expansion frame and can overwrite opacity you never changed. If nothing was stored, return without writing.

src/components/popup/message_stacks.js
```diff
@@ -18,6 +18,7 @@ export const PopupBlurMessageStacks = class PopupBlurMessageStacks {
         this.group_connections = new Map();
         this.update_ids = new Map();
         this.original_opacity = new WeakMap();
+        this.original_message_opacity = new WeakMap();
```
krissedout, line 21: original_message_opacity is introduced here, but disable() still only resets original_opacity. Please also clear original_message_opacity in disable() (same pattern as original_opacity).

krissedout (changes requested):

rinzler69-wastaken:
Thanks for catching those! I've updated restore_message_opacity to return early when no original opacity was stored, and added original_message_opacity cleanup to disable(). Pushed in 55d72f5!

krissedout (approved):

PR #956, opened 2026-08-06
Additional Popup blur fixes

### LINKED ISSUE #953: Keyboard input selection box flashing
After enabling pop-up blur, the selection box flashes when typing on the keyboard. It also flashes when selecting options from the left or right.

MeguMario:
This is mostly just cleaning up the mess I've made in previous PR

- Added `candidate-popup-content` and `candidate-popup-boxpointer` to `is_heavy_surface()` to fix https://github.com/aunetx/blur-my-shell/issues/953. This is mostly tested with ibus-mozc so for other IME, millage may vary

https://github.com/user-attachments/assets/c0408575-da0c-4aae-bd63-19439d131ec2

- Removed the old ancestor check path since skipping descendant transition scans for QS help more here. Also fixed Window switcher (alt+tab) and monitor mode popup not being blurred
- Also skipping descendant transition scans for `datemenu-popover` as well. 

- `get_monitor_clipped_surface_geometry()` now return `rect` to allow the blur to properly overflow outside of current monitor without being stop at the monitor boundary
image

https://github.com/user-attachments/assets/2020e4fc-bc4b-42db-a11f-abcc3d49e518

- Copy dark stylesheet to transparent stylesheet to at least make Transparent override look less unfinished & improve visibility as well 
image

### DIFF
--- a/src/components/popup/surface_placement.js
+++ b/src/components/popup/surface_placement.js
@@ -139,10 +139,10 @@ export const PopupBlurSurfacePlacement = class PopupBlurSurfacePlacement {
         this.monitor_index = match.monitor_index;
 
         const surface_geometry = this.create_surface_geometry(
-            match.intersection.x,
-            match.intersection.y,
-            match.intersection.width,
-            match.intersection.height
+            rect.x,
+            rect.y,
+            rect.width,
+            rect.height
         );
         surface_geometry.monitor_index = match.monitor_index;
         return surface_geometry;
--- a/src/components/popup/blur_surface.js
+++ b/src/components/popup/blur_surface.js
@@ -157,13 +157,13 @@ export const PopupBlurSurface = class PopupBlurSurface {
     
 
     is_quick_settings() {
-        return this.style.has_any_style_class(this.target, ['quick-toggle-menu','quick-settings'])
-            || this.style.has_any_style_class(this.root_actor, ['quick-toggle-menu','quick-settings']);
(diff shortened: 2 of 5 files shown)


src/components/popup/surface_placement.js
```diff
@@ -187,7 +187,7 @@ export const PopupBlurSurfacePlacement = class PopupBlurSurfacePlacement {
 
         return {
             monitor_index,
-            intersection,
+            intersection: rect,
```
krissedout: Returning intersection: rect is fine for allowing blur past the monitor edge, but find_best_monitor_intersection still ranks by match.intersection.width * height. Every overlapping monitor now gets the same area (the full rect), so monitor_index becomes the first overlap instead of the largest one. Rank using the real intersect area; only use rect as the returned geometry.

krissedout (changes requested):

src/components/popup/surface_placement.js
MeguMario: `get_monitor_clipped_surface_geometry()` now use the original `rect` instead of returning it in `get_monitor_intersection()`

krissedout (approved):
lgtm

note: the past-monitor-edge overflow is mainly a dynamic-blur win; static blur is still one-monitor-bound, as before, this isn't blocking obviously.

MeguMario:
Yeah unfortunately this is limitation of static blur that we have to live with for now. I imagine there are solutions but that would mean a big overhaul and I rather seeing shared blur finished than doing that (lol)

rinzler69-wastaken:
WACK Shell is that you?

PR #959, opened 2026-08-09
Add option to preserve shell theme for popup backgrounds

AdityaHebballe:
## Summary

- add a **Preserve shell theme** toggle under Popup Blur's background override
Screenshot From 2026-08-10 00-29-12

- keep Blur My Shell's popup surface transparency/tint while allowing the active shell theme to style popup controls and text
- share the existing canonical surface rules across Transparent, Light, and Dark modes
- preserve notification and stacked-notification surface opacity without overriding their theme-provided controls

## Motivation

Popup background overrides currently style both the blurred surface and its child controls. This can replace colors and states supplied by custom shell themes.

The new option applies a surface-only style class. Backgrounds, borders, shadows, and corner geometry still support the blur, while buttons, toggles, sliders, calendars, notification controls, and text retain the shell theme's styling.

## Before and after

| Before | After |
| --- | --- |
| before|after|

## Testing

- tested on GNOME Shell 51 with the Marble shell theme
- verified Transparent mode with Preserve shell theme enabled and disabled
- verified popup menus, Quick Settings, notifications, stacked notifications, OSDs, switchers, and dialogs have surface-rule coverage
- verified the existing non-preserve selectors retain their original declarations
- verified every surface-only selector uses declarations from a canonical existing background selector

### DIFF
--- a/resources/ui/popup.ui
+++ b/resources/ui/popup.ui
@@ -184,6 +184,20 @@
 
             <child>
               <object class="AdwActionRow">
+                <property name="title" translatable="yes">Preserve shell theme</property>
+                <property name="subtitle" translatable="yes">Keep your shell theme's colors for popup controls.</property>
+                <property name="activatable-widget">preserve_shell_theme</property>
+
+                <child>
+                  <object class="GtkSwitch" id="preserve_shell_theme">
+                    <property name="valign">center</property>
+                  </object>
+                </child>
+              </object>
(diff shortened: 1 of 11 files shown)


MeguMario (approved):
lgtm, I will leave it here for couple hours just in case you want to make some changes

krissedout (approved):
lgtm too

PR #962, opened 2026-08-15
Update Liquid Glass effect with Liquid (Gl)ass 0.1.0b shader

treebarkbr:
## Summary

Updates the Liquid Glass refraction effect with the shader features from winaviation's Liquid (Gl)ass tweak 0.1.0b, building on the earlier refraction work. The changes are ported from `LiquidAssBackboardd/Tweak.mm` (Metal) to GLSL for Blur my Shell.

## How it works

The effect already samples the pipeline texture behind the actor and uses a rounded-rect distance field to find the glass edge, corner radius, and outward normal. This change uses that edge data to drive the 0.1.0b features:

- Physical Schlick fresnel rim glare, weighted by backdrop luminance so brighter areas catch the edge a bit more naturally. This replaces the old fixed highlight lobes and the LCH re-lighting
- Per-channel chromatic dispersion with real refractive indices (R 0.98 / G 1.00 / B 1.02), so the color separation follows the refraction profile instead of a fixed offset
- Analytic superellipse and rounded-box SDF corners with a convex squircle profile, replacing the old distance-field marching
- A tint color (RGBA) blended over the backdrop, weighted by the existing tint strength
- A backdrop zoom option, plus an alpha fallback so transparent edge samples fall back to a solid tint instead of garbage
- An early-out for the flat interior of rounded surfaces, skipping refraction where it is not visible

## Known issues

- Chromatic dispersion and the alpha fallback add up to 5 backdrop texture fetches per fragment in the bezel band. This can be felt on integrated GPUs. The flat interior early-out helps, but the edge band is still the hot path
- Not ported from 0.1.0b: glyph mask refraction (iOS clock and icons), and the coversheet mode, which refracts the lock screen around the edges when you swipe up to unlock. Neither maps to a surface Blur my Shell has

## Testing

- Tested manually on GNOME 50.3, Wayland
- Checked the shader with `glslangValidator`

### DIFF
--- a/src/effects/effects.js
+++ b/src/effects/effects.js
@@ -435,7 +435,7 @@ export function get_supported_effects(_ = () => "") {
                     name: _("Tint color"),
                     description: _("Color blended over the blurred texture, weighted by the tint strength."),
                     type: "rgba",
-                    use_alpha: false
+                    use_alpha: true
                 },
                 backdrop_zoom: {
                     name: _("Backdrop zoom"),
--- a/src/effects/effects.js
+++ b/src/effects/effects.js
@@ -412,8 +412,8 @@ export function get_supported_effects(_ = () => "") {
                     digits: 0
                 },
                 gloss: {
-                    name: _("Gloss"),
-                    description: _("Strength of the white rim and surface highlight."),
+                    name: _("Fresnel glare"),
+                    description: _("Strength of the Schlick fresnel rim glare from the glass edge."),
                     type: "float",
                     min: 0.,
                     max: 1.,
@@ -422,11 +422,27 @@ export function get_supported_effects(_ = () => "") {
                     digits: 2
                 },
                 tint: {
-                    name: _("Tint"),
+                    name: _("Tint strength"),
                     description: _("Amount of subtle milky glass tint over the blurred texture."),
                     type: "float",
                     min: 0.,
                     max: 1.,
+                    increment: 0.01,
+                    big_increment: 0.1,
+                    digits: 2
+                },
+                tint_color: {
+                    name: _("Tint color"),
+                    description: _("Color blended over the blurred texture, weighted by the tint strength."),
+                    type: "rgba",
+                    use_alpha: false
+                },
+                backdrop_zoom: {
+                    name: _("Backdrop zoom"),
+                    description: _("Zoom level of the backdrop visible through the glass."),
+                    type: "float",
+                    min: 0.1,
+                    max: 4.,
                     increment: 0.01,
                     big_increment: 0.1,
                     digits: 2
@@ -442,8 +458,8 @@ export function get_supported_effects(_ = () => "") {
                     digits: 2
                 },
                 rgb_fringing: {
-                    name: _("RGB fringing"),
-                    description: _("Adds subtle color separation to the bent texture."),
+                    name: _("Chromatic dispersion"),
+                    description: _("Physical per-channel refraction dispersion at the glass edge."),
                     type: "float",
                     min: 0.,
                     max: 1.,
--- a/src/effects/refraction.glsl
+++ b/src/effects/refraction.glsl
@@ -1,4 +1,4 @@
-// GLSL port of winaviation-tweaks/liquidass Shared/LGMetalShaderSource.m.
+// GLSL port of winaviation-tweaks/liquidass 0.1.0b LiquidAssBackboardd/Tweak.mm kShaderSrc.
 // The sampling and blur pass are adapted for Blur my Shell's Clutter pipeline.
 uniform sampler2D tex;
 uniform float width;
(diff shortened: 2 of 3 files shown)


src/effects/effects.js
```diff
@@ -431,6 +431,22 @@ export function get_supported_effects(_ = () => "") {
                     big_increment: 0.1,
                     digits: 2
                 },
+                tint_color: {
+                    name: _("Tint color"),
+                    description: _("Color blended over the blurred texture, weighted by the tint strength."),
+                    type: "rgba",
+                    use_alpha: false
```
krissedout: use_alpha: false makes the shared color picker save this value as [r, g, b], but pack_param() only accepts four-component color arrays, changing the tint color therefore throws an "unsupported type" error and prevents the pipeline settings from being saved. please preserve/store an alpha component even when the picker does not expose alpha.

src/effects/refraction.glsl
```diff
@@ -227,32 +183,103 @@ struct EdgeInfo {
...
+
+    vec4 fallback = vec4(0.0);
+    bool loadedFallback = false;
+    if (greenSample.a < 0.01) {
+        fallback = sampleGlassBackdrop(resolveUV(sampleUV));
+        loadedFallback = true;
+        greenSample = fallback;
+    }
+    if (greenSample.a < 0.01)
+        return vec4(0.0);
```
krissedout: this does not provide the solid-tint fallback described by the PR. if both the displaced sample and the undisplaced fallback are transparent, the shader returns transparent black here, and main() exits before applying the configured tint. please return or construct a tint-backed sample for this case.

src/effects/refraction.glsl
```diff
@@ -316,55 +343,45 @@ void main() {
...
         edgeBand = clamp(1.0 - (distFromSide / rimRadius), 0.0, 1.0);
     }
 
+    // 0.1.0b early-out: away from the edge, just a tinted flat sample.
+    // (skipped for the circular surface, which refracts across the whole face)
+    if (!useCircularSurface && R < shortestSide * 0.45 && distFromSide >= bezel) {
+        vec4 flatSample = sampleGlassBackdrop(actorUV);
+        flatSample.rgb = mix(flatSample.rgb, vec3(tint_r, tint_g, tint_b), tint * tint_a);
+        flatSample.rgb *= 1.0 - smoothstep(0.25, 1.0, localUV.y) * shadow * 0.20;
+        cogl_color_out = vec4(clamp(flatSample.rgb, 0.0, 1.0) * edgeOpacity, edgeOpacity * opacity_factor);
```
krissedout: this early-out bypasses backdropSampleUV(), so backdrop_zoom affects only the bezel for typical rounded rectangles while the interior remains unzoomed, with any zoom other than 1.0, that also creates a sampling discontinuity at distFromSide == bezel. the flat sample should use the same zoom transform with zero displacement.

krissedout (changes requested):

treebarkbr:
Addressed.

src/effects/refraction.glsl
```diff
@@ -316,55 +343,46 @@ void main() {
...
+        return;
+    }
+
     float bezelRatio = useCircularSurface
         ? clamp(distFromSide / lensBezel, 0.0, 1.0)
         : clamp(distFromSide / max(1.0, bezel * 0.35 * max(1.0, rim_width)), 0.0, 1.0);
-    float normDisp = edgeBand > 0.001
+    float normDisp = distFromSide < bezel
```
krissedout: the cutoff still does not match the rim_width scaled refraction band. with the default rim_width of 4.8, rimRadius is 1.68 * bezel, so displacement remains nonzero immediately before distFromSide == bezel and then abruptly becomes zero at this early-out/normDisp cutoff. at higher supported settings this can be a multi-pixel jump. please use one consistent band width for edgeBand, bezelRatio, the early-out, and the displacement cutoff.

src/effects/refraction.glsl
```diff
@@ -227,32 +183,103 @@ struct EdgeInfo {
...
+        vec4 blueSample = sampleGlassBackdrop(blueUV);
+
+        if (redSample.a < 0.01 || blueSample.a < 0.01) {
+            if (!loadedFallback)
+                fallback = sampleGlassBackdrop(resolveUV(sampleUV));
+            if (redSample.a < 0.01)
+                redSample = fallback;
+            if (blueSample.a < 0.01)
+                blueSample = fallback;
+        }
```
krissedout, line 294: the solid-tint fallback is still incomplete for the dispersed red and blue samples. if one channel sample is transparent and the undisplaced fallback is also transparent, that transparent sample is assigned unchanged and its zero RGB channel is copied into bg, producing an incorrect color fringe. please substitute the corresponding tint channel when the fallback remains transparent, as the green path now does.

krissedout (changes requested):

krissedout (approved):

PR #963, opened 2026-08-19
perf: throttle shadow-driven blur repaints

Dank-del:
Replaces frame-count-based blur invalidation with a monotonic 100 ms throttle. This keeps shadow-driven blur refreshes responsive while preventing continuously painting actors from requeuing costly blur work every few frames. Built successfully with the project Makefile.

### DIFF
--- a/src/conveniences/paint_signals.js
+++ b/src/conveniences/paint_signals.js
@@ -16,24 +16,48 @@ export const PaintSignals = class PaintSignals {
         this.disconnect_all_for_actor(actor);
 
         let last_repaint_at = 0;
-        const paint_effect = new PaintCallbackEffect();
-        paint_effect.set_callback(() => {
-            const now = GLib.get_monotonic_time();
-            if (now - last_repaint_at < REPAINT_INTERVAL_US)
-                return;
-
-            last_repaint_at = now;
+        let repaint_timeout_id = 0;
+        const queue_repaint = () => {
+            last_repaint_at = GLib.get_monotonic_time();
             try {
                 blur_effect.queue_repaint();
             } catch (e) { }
+        };
+        const cancel_repaint = () => {
+            if (!repaint_timeout_id)
+                return;
+
+            GLib.Source.remove(repaint_timeout_id);
+            repaint_timeout_id = 0;
+        };
+        const paint_effect = new PaintCallbackEffect();
+        paint_effect.set_callback(() => {
+            const now = GLib.get_monotonic_time();
+            const elapsed = now - last_repaint_at;
+            if (elapsed >= REPAINT_INTERVAL_US) {
+                cancel_repaint();
+                queue_repaint();
+                return;
+            }
+
+            if (repaint_timeout_id)
+                return;
+
+            const delay = Math.max(1, Math.ceil((REPAINT_INTERVAL_US - elapsed) / 1000));
+            repaint_timeout_id = GLib.timeout_add(GLib.PRIORITY_DEFAULT, delay, () => {
+                repaint_timeout_id = 0;
+                queue_repaint();
+                return GLib.SOURCE_REMOVE;
+            });
         });
 
         actor.add_effect(paint_effect);
         const destroy_id = this.connections.connect(actor, 'destroy', () => {
             paint_effect.set_callback(null);
+            cancel_repaint();
             this.entries.delete(actor);
         });
-        this.entries.set(actor, { paint_effect, destroy_id });
+        this.entries.set(actor, { paint_effect, destroy_id, cancel_repaint });
     }
 
     disconnect_all_for_actor(actor) {
@@ -43,6 +67,7 @@ export const PaintSignals = class PaintSignals {
 
         this.entries.delete(actor);
         entry.paint_effect.set_callback(null);
+        entry.cancel_repaint();
         this.connections.disconnect(actor, entry.destroy_id);
         try {
             actor.remove_effect(entry.paint_effect);
--- a/src/conveniences/paint_signals.js
+++ b/src/conveniences/paint_signals.js
@@ -32,10 +32,7 @@ export const PaintSignals = class PaintSignals {
         };
         const paint_effect = new PaintCallbackEffect();
         paint_effect.set_callback(paint_flags => {
-            // queue_repaint() invalidates the effect itself, which results in
-            // a paint without ACTOR_DIRTY. Do not turn that repaint into a
-            // new trailing timeout.
-            if (!(paint_flags & Clutter.PaintFlag.ACTOR_DIRTY))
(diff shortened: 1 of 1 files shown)


src/conveniences/paint_signals.js
```diff
@@ -12,15 +15,14 @@ export const PaintSignals = class PaintSignals {
...
-        let paints_to_skip = 0;
+        let last_repaint_at = 0;
         const paint_effect = new PaintCallbackEffect();
         paint_effect.set_callback(() => {
-            if (paints_to_skip > 0) {
-                paints_to_skip--;
+            const now = GLib.get_monotonic_time();
+            if (now - last_repaint_at < REPAINT_INTERVAL_US)
                 return;
-            }
```
krissedout, line 21: This throttle drops paints inside the 100 ms window without scheduling a trailing invalidation. For example, at 60 Hz a repaint at t=0 followed by paints at t=16/32/48 ms will suppress every update; if the animation then stops, the blur remains stuck at the t=0 state indefinitely. The previous counter would repaint on that fourth frame. Please coalesce suppressed paints into one trailing repaint after the remaining interval, and cancel that timer when the actor is disconnected or destroyed.

krissedout (changes requested):

src/conveniences/paint_signals.js
```diff
@@ -12,26 +15,49 @@ export const PaintSignals = class PaintSignals {
...
-            } catch (e) { }
+            if (repaint_timeout_id)
+                return;
+
+            const delay = Math.max(1, Math.ceil((REPAINT_INTERVAL_US - elapsed) / 1000));
+            repaint_timeout_id = GLib.timeout_add(GLib.PRIORITY_DEFAULT, delay, () => {
+                repaint_timeout_id = 0;
+                queue_repaint();
+                return GLib.SOURCE_REMOVE;
+            });
```
krissedout, line 63: The trailing timeout can create a permanent 10 Hz repaint loop. queue_repaint() causes another effect paint; that paint arrives inside the throttle window with no pending timeout and schedules the next timeout, which repeats indefinitely even after the original actor updates stop. Clutter marks effect-initiated paints by omitting ACTOR_DIRTY, so please pass paint_flags into the callback and ignore self-initiated paints, or otherwise suppress the paint caused by queue_repaint().

krissedout (changes requested):

src/conveniences/paint_signals.js
```diff
@@ -12,26 +15,55 @@ export const PaintSignals = class PaintSignals {
...
         const paint_effect = new PaintCallbackEffect();
-        paint_effect.set_callback(() => {
-            if (paints_to_skip > 0) {
-                paints_to_skip--;
+        paint_effect.set_callback(paint_flags => {
+            // queue_repaint() invalidates the effect itself, which results in
+            // a paint without ACTOR_DIRTY. Do not turn that repaint into a
+            // new trailing timeout.
+            if (!(paint_flags & Clutter.PaintFlag.ACTOR_DIRTY))
+                return;
```
krissedout: This uses the wrong Clutter enum. The vfunc_paint() argument is Clutter.EffectPaintFlags, while Clutter.PaintFlag is a separate stage-paint enum and has no ACTOR_DIRTY member. Consequently, undefined is coerced to zero here and every paint returns early, disabling blur invalidation entirely. Please use Clutter.EffectPaintFlags.ACTOR_DIRTY. [Clutter documents the callback flag as ClutterEffectPaintFlags](https://gnome.pages.gitlab.gnome.org/mutter/clutter/class.ShaderEffect.html#clutter_effect_queue_repaint)

krissedout (changes requested):

src/conveniences/paint_signals.js
```diff
@@ -12,26 +15,52 @@ export const PaintSignals = class PaintSignals {
...
+            GLib.Source.remove(repaint_timeout_id);
+            repaint_timeout_id = 0;
+        };
         const paint_effect = new PaintCallbackEffect();
-        paint_effect.set_callback(() => {
-            if (paints_to_skip > 0) {
-                paints_to_skip--;
+        paint_effect.set_callback(paint_flags => {
+            if (!(paint_flags & Clutter.EffectPaintFlags.ACTOR_DIRTY))
+                return;
```
krissedout, line 45: Sorry, it looks like I made a mistake in my previous suggestion 🫣. ACTOR_DIRTY cannot distinguish the repaint triggered by blur_effect.queue_repaint() here. Clutter clears this flag only for the effect passed to queue_repaint; effects layered above it still receive ACTOR_DIRTY. Since PaintCallbackEffect is added after the blur effect, each timeout-triggered repaint reaches this callback and schedules another timeout, preserving the permanent 10 Hz repaint loop. This needs an explicit one-shot guard for the requested repaint instead of relying on paint flags.

krissedout (changes requested):

aunetx:
Hi, sorry for not having been able to answer anything during the past months. This is just to tell you @krissedout and @MeguMario that it looks like you merged quite a lot of PRs recently; I don't know at which state the project is but if you need me to make a release on extensions.gnome.org don't hesitate to ping me here or send me an email!

Thanks for your work :D

MeguMario:
@aunetx Hi there, good to see you drop by. 

Regarding the state of the project, I think it's in a stable enough state at least, there were time that critical bugs drip down to users (those of ublue's) but at least those should be solved by now (also Bazzite ship the rounded blur library by default now). 

As for making new releases, personally I think once this PR is merged, no new problem was found and if @krissedout give an okay sign, it's new release time (it's about time for GNOME 51 after all). 

Imma wake up @Dank-del at least

src/conveniences/paint_signals.js
Dank-del: Fixed in a30e637 by adding explicit one-shot suppression around queue_repaint(). The callback consumes the guard before checking paint flags, so the requested repaint cannot schedule another timeout; the guard is reset if queue_repaint() throws.

Dank-del:
@krissedout Addressed the remaining repaint-loop issue in a30e637. Could you take another look when you have a chance?

krissedout:
@aunetx @MeguMario I agree with the release, I think it's stable enough and too many changes have been made so we should release soon, and before the big refactor comes, also from @aunetx I need you to unlock weblate as it's been locked since June 30th so people have no way of contributing translations!

krissedout (approved):

MeguMario:
@aunetx @krissedout A little message to say that we should probably wait until #974 fixed to make a new release. Sorry for the trouble

aunetx:
Thanks for your answers @MeguMario @krissedout. I unlocked weblate, and will wait for your ping for making a release then!

PR #965, opened 2026-08-20
Add the option to hide blur while an app is fullscreen

### LINKED ISSUE #703: Weird behavior in overview
the blur behind hovered window in overview doesn't render properly as shown in the recording below , can you help me get it right ?
[Screencast From 2025-01-05 20-39-37.webm](https://github.com/user-attachments/assets/ea46607f-5e8a-4048-b6e4-c9d0bf1a1df1)

### LINKED ISSUE #749: how to disable application blur in overview ?
HI,
i am a good fan of your work on this extension with unique feature that bring more life to my workflow.

I really appreciate the feature that blurs application windows. However, when I use CoverFlow or when I switch to Overview mode, I often regret that the blurring is always active. This isn't a big problem when there are one or two or three windows, but when there are more, visibility becomes reduced, and sometimes a little disturbing.

I particularly appreciate the blurring of windows on the desktop, but not really in Overview mode or with CoverFlow. How can I disable the blurring in both cases?

### LINKED ISSUE #899: Allow disabling blur when  on fullscreen
###  Issue:
When on full screen mode (watching or gaming), the blur effect is noticeable. 

### Request:
Add a toggle that can disable the blur effect when the app is on full screen.

MeguMario:
This PR add the ability to hide blur while an app is in fullscreen (https://github.com/aunetx/blur-my-shell/issues/899), plus some changes which I'm going to list here
image
(probably will also set the opacity to 255 as well)

- Application blur will only render on overview if Blur on overview is on (https://github.com/aunetx/blur-my-shell/issues/749)

### DIFF
--- a/src/components/applications.js
+++ b/src/components/applications.js
@@ -173,6 +173,39 @@ export const ApplicationsBlur = class ApplicationsBlur {
                 }
             );
         }
+        else {
+            this.connections.connect(
+                Main.overview, 'showing',
+                _ => this.meta_window_map.forEach((meta_window, _pid) => {
+                    let window_actor = meta_window.get_compositor_private();
+                    window_actor?.hide();
+                })
+            );
+            // when the overview is closed, hide every actor that is not on the
+            // current workspace (to mimic the original behaviour)
+            this.connections.connect(
+                Main.overview, 'hidden',
+                _ => { this.meta_window_map.forEach((meta_window, _pid) => {
+                        let window_actor = meta_window.get_compositor_private();
+                        if (
+                            (!meta_window.get_workspace().active) || meta_window.minimized
+                        ){
+                            window_actor.hide();
+                        }
+                        else {
+                            window_actor.show();
+                            if (this.settings.applications.UNBLUR_WHEN_FULLSCREEN) {
+                                 this.unblur_when_fullscreen(meta_window);
+                            }     
+                            else {
+                                meta_window.blur_actor?.show();
+                            }
+                                
+                        }
+                    });
+                }
+            );
+        }
     }
 
     /// Iterate through all existing windows and add blur as needed.
@@ -375,6 +408,9 @@ export const ApplicationsBlur = class ApplicationsBlur {
 
         meta_window.blur_actor = blur_actor;
 
+        if (!this.settings.applications.BLUR_ON_OVERVIEW && Main.overview.visible)
+            blur_actor.hide();
+
         // make sure window is blurred in overview
         if (this.settings.applications.BLUR_ON_OVERVIEW)
             this.enforce_window_visibility_on_overview_for(window_actor);
@@ -387,6 +423,11 @@ export const ApplicationsBlur = class ApplicationsBlur {
 
         // update corner radius based on window state
         this.update_corner_radius(meta_window);
+
+        // unblur when fullscreen
+        if (this.settings.applications.UNBLUR_WHEN_FULLSCREEN) {
+            this.unblur_when_fullscreen(meta_window);
+        }
 
         // now set up the signals, for the window actor only: they are disconnected
         // in `remove_blur`, whereas the signals for the meta window are disconnected
@@ -403,7 +444,12 @@ export const ApplicationsBlur = class ApplicationsBlur {
         );
         this.connections.connect(
             meta_window, 'notify::fullscreen',
-            _ => this.update_corner_radius(meta_window)
+            _ => {
+                this.update_corner_radius(meta_window);
+                if (this.settings.applications.UNBLUR_WHEN_FULLSCREEN){
+                    this.unblur_when_fullscreen(meta_window);
+                }
+            }
         );
 
         // update the window opacity when it changes, else we don't control it fully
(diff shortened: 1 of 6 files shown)


src/components/applications.js
```diff
@@ -509,13 +568,38 @@ export const ApplicationsBlur = class ApplicationsBlur {
...
+            window_actor = meta_window.get_compositor_private();
+        }
+
+        if (is_fullscreen) {
+            meta_window.blur_actor.hide();
+            this.set_window_opacity(window_actor, 255);
+        } else {
+            meta_window.blur_actor.show();
+            this.set_window_opacity(window_actor, this.settings.applications.OPACITY);
+        }
```
krissedout: This restoration ignores the other visibility rules. When a focused window exits fullscreen, this unconditionally shows its blur and applies the configured opacity even though dynamic opacity requires the focused window to remain opaque with its blur hidden. Overview visibility has the same conflict. Please centralize fullscreen, focus, overview, and actor visibility into one state reconciliation method.

src/components/applications.js
```diff
@@ -509,13 +568,38 @@ export const ApplicationsBlur = class ApplicationsBlur {
...
             (meta_window, _pid) => this.update_corner_radius(meta_window)
         )
     }
 
+    update_fullscreen_status() {
+        if (this.settings.applications.UNBLUR_WHEN_FULLSCREEN){
+            this.meta_window_map.forEach(
+                (meta_window, _pid) => this.unblur_when_fullscreen(meta_window)
+            )
+        }
```
krissedout: This only updates windows when the setting is enabled. Turning the option off leaves currently fullscreen windows at opacity 255 with their blur hidden, and notify::fullscreen will not fire again to restore them. The setting callback needs to reconcile every tracked window for both values.

krissedout (changes requested):
I think you should split the application behavior from the unrelated popup and dock stylesheet changes. The fullscreen and overview changes currently scatter competing show(), hide(), and opacity updates across several callbacks. They need one authoritative state reconciler before this is safe to merge

MeguMario:
@krissedout styling changes have been splited off to https://github.com/aunetx/blur-my-shell/pull/970

src/components/applications.js
MeguMario: the logic has now been centralized in `reconcile_window_visibility()`

src/components/applications.js
```diff
@@ -509,13 +530,44 @@ export const ApplicationsBlur = class ApplicationsBlur {
...
+            && (this.settings.applications.BLUR_ON_OVERVIEW || !Main.overview.visible);
+
+        if (show_blur)
+            blur_actor.show();
+        else
+            blur_actor.hide();
+
+        this.set_window_opacity(window_actor, is_focused || is_fullscreen
+            ? 255
+            : this.settings.applications.OPACITY);
```
krissedout: When overview blur is disabled, show_blur becomes false, but this still applies the configured opacity unless the window is focused or fullscreen. With the default opacity of 215, overview clones are left translucent after their blur has been hidden. The configured opacity should only be used when show_blur is true; every state that suppresses blur should restore 255.

src/components/applications.js
```diff
@@ -167,12 +168,49 @@ export const ApplicationsBlur = class ApplicationsBlur {
...
+                            window_actor.hide();
+                        }
+                        else {
+                            window_actor.show();
+                        }
+
+                        this.reconcile_window_visibility(meta_window);
+                    });
+                }
+            );
```
krissedout: This branch still takes ownership of MetaWindowActor visibility: it force-hides every tracked window when overview opens and force-shows every active, non-minimized window when it closes. That can override visibility state owned by Shell which this code does not model. These callbacks should only trigger the blur/opacity reconciler; leave window_actor.show() and hide() to Shell.

krissedout (changes requested):
Thanks for splitting this and centralizing the application state! The original fullscreen/focus conflicts are resolved, but there's a few more issues.

src/components/applications.js
```diff
@@ -509,13 +481,44 @@ export const ApplicationsBlur = class ApplicationsBlur {
...
+            return;
+
+        const is_focused = this.settings.applications.DYNAMIC_OPACITY
+            && meta_window.bms_pid === this.focused_window_pid;
+        const is_fullscreen = this.settings.applications.UNBLUR_WHEN_FULLSCREEN
+            && meta_window.fullscreen;
+        const show_blur = window_actor.visible
+            && !is_focused
+            && !is_fullscreen
+            && (this.settings.applications.BLUR_ON_OVERVIEW || !overview_visible);
```
krissedout: Tracking overview state explicitly fixes the signal timing, but this predicate still excludes minimized and inactive-workspace windows because their MetaWindowActor is hidden even though Shell paints it through a ClutterClone in overview. Treat the actor as paint-visible when BLUR_ON_OVERVIEW && overview_visible, then remove the old enforce_window_visibility_on_overview_for() actor-forcing hook. We should solve this entirely in the reconciler, not restore the old show()/hide() callbacks.

Something like:
```
const visible_for_blur = window_actor.visible
    || (this.settings.applications.BLUR_ON_OVERVIEW && overview_visible);
```

krissedout (changes requested):

krissedout (approved):

PR #967, opened 2026-08-25
feat(applications): animate the applications blur with the window

### LINKED ISSUE #649: Blur disappears when moving a window with Compiz enabled
Only 2 extensions involved are BMS and Compiz window effect. BMS works fine alone, but with enabled Compiz, whenever the animation is active the BMS transparency stays on, but blur disappears. Couldn't find any related settings in Compiz menu.

NB-Group:
Fixes #649.

When a window is minimized or unminimized through an animation effect (compiz-alike-magic-lamp-effect here), the applications blur disappears for the duration of the animation. This keeps the blur alive and makes it follow the window through the animation.

Demo, a photo of my screen taken with a phone, since the stock screenshot tool cannot capture a window mid-animation:

demo

The feature is experimental and opt-in, through a "Blur during animations (experimental)" switch in the applications preferences (`animation-blur` key), with an "Animation blur brightness" slider next to it. The gaussian effect reads brighter than the native one at the same nominal value, so the slider compensates (default 0.8, tuned by eye).

How it works, while an animation runs a sampler actor sits below the animating window:

- once per frame the background behind the window is captured with `stage.paint_to_content` (the window, the sampler and the blur widget are hidden during the capture so nothing feeds back into itself) and blurred with the existing GaussianBlurEffect, so the blurred content stays flat and live while only the outline moves;
- a mask shader analytically inverts the magic-lamp deform and clips the flat blur to the animated silhouette, reusing the resting blur's corner radius and frame insets (CSD shadow);
- the resting blur widget is hidden while the animation runs, and both are handed back in a single idle callback so the switch happens in one frame. The destroy is kept out of the timeline callback because destroying a Clutter actor there during a GC sweep crashes the shell.

Known limitations:

- the animation blur only runs for a bottom dock; for the other dock positions the deform is not inverted yet and the vanilla behaviour is kept;
- the per-frame capture costs one offscreen stage paint per frame while an animation runs. Fine on my machine (GNOME 50, NVIDIA, fractional scaling), but it is the thing to look at if someone reports stutter.


krissedout:
Thanks for the work here, the prototype is clever, but I don’t think we should merge this implementation.

This adds a second animation-only renderer that captures and blurs the stage every frame instead of preserving the existing application blur. It doesn’t reuse the configured pipeline, requires separate visual tuning, depends on Magic Lamp’s private implementation details and bottom-dock geometry, and adds nearly 500 lines to an already oversized applications.js. Even with the compatibility and lifecycle bugs fixed, we’d be maintaining two renderers that can drift visually and behaviorally.

More fundamentally, ext-background-effect-v1 is landing in GNOME 51, allowing Wayland applications to request compositor-owned background blur for their own surfaces. That is where application blur should live going forward. If you want blur in a particular application, please open an issue with that application or its toolkit to implement the protocol instead of adding more app-specific rendering machinery to BMS.

NB-Group:
Thank you for your review and response. I completely agree with your viewpoint. The "ext-background-effect-v1" feature is excellent, and I look forward to its implementation.

PR #968, opened 2026-08-25
Support blurring screenshot-ui-panel for popup blur

MeguMario:
This PR add support for blurring `screenshot-ui-panel` for Popup blur

Demo video for `screenshot-ui-panel` 

- Dynamic blur

https://github.com/user-attachments/assets/fdd4da72-934d-44d6-b630-09d94e92f091

- Static blur

https://github.com/user-attachments/assets/dddab027-6f98-4512-8406-a83a2fcbc7fb

I originally included this with #965 but decided that it's gotten too big and split this off into its own PR instead

PR #164, opened 2022-01-07
Blur onscreen keyboard

aunetx:
This is an early attempt to blur the onscreen keyboard of gnome, it has problems when first launching it (no blur first time).

PR #971, opened 2026-08-27
Support blurring OSK for popup blur

### LINKED ISSUE #846: [feature] Blur under original On Screen Keyboard
It would be nice if we had an blur like this imho (not exactly like this, but something similar)
Image
_buttons are ugly_

MeguMario:
A spiritual successor to https://github.com/aunetx/blur-my-shell/pull/164

https://github.com/user-attachments/assets/be047ed7-61d5-4fcc-9f4e-4293922ac8a9

Splited off from https://github.com/aunetx/blur-my-shell/pull/968

### DIFF
--- a/src/components/popup/index.js
+++ b/src/components/popup/index.js
@@ -14,6 +14,7 @@ import { PopupBlurMessageStacks } from './message_stacks.js';
 
 const POPUP_INTERNAL_STYLE_CLASSES = ['bms-popup-blurred-widget', 'bms-popup-backgroundgroup'];
 const POPUP_INTERNAL_NAMES = ['bms-popup-blurred-widget', 'bms-popup-backgroundgroup'];
+const KEYBOARD_STYLE_CLASS = 'bms-keyboard-surface';
 
 export const PopupBlur = class PopupBlur {
     constructor(connections, settings, effects_manager) {
@@ -56,7 +57,6 @@ export const PopupBlur = class PopupBlur {
             'changed::gtk-theme',
             () => this.update_background()
         );
-        
         this.connections.connect(
             Main.sessionMode,
             'updated',
@@ -76,6 +76,8 @@ export const PopupBlur = class PopupBlur {
         this.track_container(Main.messageTray?._bannerBin);
         this.track_quick_settings();
         this.track_container(global.window_group);
+        this.track_keyboard();
+        this.track_container(Main.keyboard?.keyboardActor);
     }
 
     track_container(container) {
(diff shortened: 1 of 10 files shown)


src/components/popup/index.js
```diff
@@ -449,11 +479,38 @@ export const PopupBlur = class PopupBlur {
...
+        this.clear_keyboard_style(actor);
+        if (!this.settings.popup.OVERRIDE_BACKGROUND)
+            return;
+
+        const style = background_style ?? (
+            this.settings.popup.PRESERVE_SHELL_THEME ?
+                POPUP_SURFACE_STYLES[this.get_background_style()] :
+                POPUP_BACKGROUND_STYLES[this.get_background_style()]
+        );
+        actor.add_style_class_name(KEYBOARD_STYLE_CLASS);
```
krissedout: bms-keyboard-surface is both the OSK target identity and optional visual styling, but this removes it whenever Override background is disabled. On initial enable that means no OSK blur surface is discovered at all; after toggling the setting, an existing surface can remain without the keyboard-specific fade handling. Keep the internal OSK target marker unconditional and only gate the background style class.

krissedout (changes requested):

src/components/popup/index.js
```diff
@@ -449,11 +479,43 @@ export const PopupBlur = class PopupBlur {
             Main.uiGroup.add_style_class_name(
                 background_style
             );
+            keyboard_actors.forEach(actor => this.update_keyboard_style(actor, background_style));
```
krissedout: This is the only call that marks existing keyboard actors, but it is gated by OVERRIDE_BACKGROUND. When an OSK already exists and background override is disabled, it never receives bms-keyboard-surface and is never discovered as a blur target. Call update_keyboard_style() for existing actors outside this branch and gate only the visual background class.

krissedout (changes requested):

src/components/popup/index.js
```diff
@@ -450,10 +480,41 @@ export const PopupBlur = class PopupBlur {
                 background_style
             );
         }
-
+        keyboard_actors.forEach(actor => this.update_keyboard_style(actor, background_style));
```
krissedout, line 484: background_style is declared with const inside the preceding if block, so this throws ReferenceError every time update_background() runs. Declare let background_style = null before the condition, assign it inside, then keep this unconditional call outside.

krissedout (changes requested):

src/components/popup/index.js
krissedout, line 479: The outer background_style is never assigned because line 477 declares another block-scoped const with the same name. Line 484 therefore always passes null, causing every keyboard actor to recompute the style. Change `const background_style =` to `background_style =`.

krissedout (changes requested):

krissedout (approved):

PR #972, opened 2026-08-27
Rounded Blur library install script update

MeguMario:
- Update the script to now inform Fedora user that they should use dnf to install the library from now on using external copr (https://github.com/ublue-os/packages/pull/1336)
- Support GNOME 51 (https://github.com/kancko/gnome-rounded-blur/pull/4)

### DIFF
--- a/scripts/GUIDE.md
+++ b/scripts/GUIDE.md
@@ -35,6 +35,31 @@ paru -S gnome-rounded-blur
 yay -S gnome-rounded-blur
 ```
 
+### For Fedora (or any of its derivaties, including Atomic based distro) users
+
+Fedora user can follow the following steps to install the libray
+
+- **REMEMBER TO UNINSTALL THE LIBRARY FIRST IF YOU ALREADY INSTALL IT VIA THE SCRIPT HERE**, you can uninstall it by following the uninstall command below
+- Enable the following copr using this command 
+```
+sudo dnf copr enable aneagle/gnome-rounded-blur
+```
+- Then, install the library using `dnf`
+```
+sudo dnf install gnome-rounded-blur
+```
+
+**Note:**
+- Fedora Atomic user may want to manually add the copr by downloading the copr `.repo` from [here](https://copr.fedorainfracloud.org/coprs/aneagle/gnome-rounded-blur/) and copy the file into `/etc/yum.repos.d/`, then use the following command to refresh the metadata
+```
+sudo rpm-ostree refresh-md --force 
+```
+- Bazzite (or any of ubluek atomic distro) can directly install the library via the following command (Fedora Atomic user after following the previous step can use this as well)
+```
+rpm-ostree install gnome-rounded-blur
+```
+
+
 ### Build it yourself
 
 You can visit the original repo [here](https://github.com/kancko/gnome-rounded-blur) for guide on how to build the library yourself. Do keep in mind that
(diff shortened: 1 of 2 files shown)


scripts/rounded_blur_build.sh
```diff
@@ -21,6 +21,30 @@ check_env(){
 		sleep 5
 		exit 1
 	fi
+
+	if [[ "$OS_ID_TYPE" = "fedora" ]] || [[ "$OS_LIKE_ID_TYPE" = "fedora" ]]; then
+		if [[ $i = "y" ]] && [[ $u = "n" ]]; then
```
krissedout: This does not detect the Fedora derivatives advertised by the guide. The grep above preserves quotes, so Bazzite’s ID_LIKE="fedora" becomes "fedora" and fails this equality check; space-separated ID_LIKE values fail too. Source `/etc/os-release` and token-match ID/ID_LIKE, otherwise derivatives fall through and the script still installs directly into /usr.

scripts/GUIDE.md
```diff
@@ -35,6 +35,31 @@ paru -S gnome-rounded-blur
 yay -S gnome-rounded-blur
 ```
 
+### For Fedora (or any of its derivaties, including Atomic based distro) users
```
krissedout: "derivaties" should be "derivatives"

scripts/GUIDE.md
```diff
@@ -35,6 +35,31 @@ paru -S gnome-rounded-blur
...
+
+Fedora user can follow the following steps to install the libray
```
krissedout: "libray" should be "library"

krissedout (changes requested):

scripts/rounded_blur_build.sh
```diff
@@ -137,35 +171,58 @@ prep_stage(){
...
+		fi
 	else
-		DIFF_VALUE=$(echo "$HARDCODE_MUTTER_SYS_VER - $MUTTER_SYS_VER" | bc)
-		DIFF_VALUE_2=$(echo "$MUTTER_API_REPO_VER - $DIFF_VALUE" | bc)
-		sed -i -e '0,/'"mutter_req = ""$HARDCODE_MUTTER_SYS_VER"'/{s/'"$HARDCODE_MUTTER_SYS_VER"'/'"$MUTTER_SYS_VER"'/g}' meson.build
-		sed -i -e '0,/'"mutter_api_version = ""$MUTTER_API_REPO_VER"'/{s/'"$MUTTER_API_REPO_VER"'/'"$DIFF_VALUE_2"'/g}' meson.build
+		if [[ "$MUTTER_SYS_VER" -ge 51 ]]; then
+			DIFF_VALUE_2="$MUTTER_SYS_VER"
+			sed -i -E "s/mutter_api_version = '[0-9]+'/mutter_api_version = '$DIFF_VALUE_2'/" meson.build
+			sed -i -E "s/mutter_req = '>= [0-9.]+'/mutter_req = '>= $MUTTER_SYS_VER.0'/" meson.build
```
krissedout: This rewrites mutter_api_version and mutter_req, but the current upstream meson.build also hardcodes libmutter_dep = dependency('libmutter-18'). GNOME 46, 49, and 51 therefore still request libmutter-18 after the other references are rewritten to 14, 17, or 51, and Meson configuration fails. Update every Mutter dependency reference or change upstream to derive them all from one variable.

krissedout (changes requested):

krissedout (approved):

PR #973, opened 2026-08-29
More popup blur style tweaks

MeguMario:
This introduce a couple more tweaks to Popup blur styling:
- Add button state styling for `.switcher-list`, `.message-expand-button` and `.message-media-control`

There's supposed to be state styling for `.message` as well, there's styling for it but they don't work on touchscreen for some reason

### DIFF
--- a/src/styles/popup/dark.css
+++ b/src/styles/popup/dark.css
@@ -10,6 +10,54 @@
     box-shadow: none;
 }
 
+.bms-popup-background-dark .message-media-control,
+.bms-popup-surface-dark .message-media-control {
+  background-color: transparent;
+}
+
+.bms-popup-background-dark .message-media-control:last-child,
+.bms-popup-surface-dark .message-media-control:last-child {
+  background-color: transparent;
+}
+
+.bms-popup-background-dark .message-media-control:first-child:last-child,
+.bms-popup-surface-dark .message-media-control:first-child:last-child {
+  background-color: transparent;
+}
+
+.bms-popup-background-dark .message-media-control:hover,
+.bms-popup-surface-dark .message-media-control:hover,
+.bms-popup-background-dark .message-expand-button:hover,
+.bms-popup-surface-dark .message-expand-button:hover,
+.bms-popup-background-dark .switcher-list .item-box:hover,
+.bms-popup-surface-dark .switcher-list .item-box:hover {
+  background-color: rgba(255, 255, 255, 0.20) !important;
+}
+
+.bms-popup-background-dark .message-media-control:active,
+.bms-popup-surface-dark .message-media-control:active,
+.bms-popup-background-dark .message-expand-button:active,
+.bms-popup-surface-dark .message-expand-button:active,
+.bms-popup-background-dark .switcher-list .item-box:active,
+.bms-popup-surface-dark .switcher-list .item-box:active {
+  background-color: rgba(255, 255, 255, 0.24) !important;
+}
+
+.bms-popup-background-dark .message-media-control:focus,
+.bms-popup-background-dark .message-media-control:checked,
+.bms-popup-surface-dark .message-media-control:focus,
+.bms-popup-surface-dark .message-media-control:checked,
+.bms-popup-background-dark .message-expand-button:focus,
+.bms-popup-background-dark .message-expand-button:checked,
+.bms-popup-surface-dark .message-expand-button:focus,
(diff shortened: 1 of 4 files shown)


src/styles/popup/dark.css
```diff
@@ -10,6 +10,61 @@
...
 
+.bms-popup-background-dark .message-media-control,
+.bms-popup-surface-dark .message-media-control {
+  background-color: transparent;
+}
+
+.bms-popup-background-dark .message-media-control:last-child:not(:first-child),
+.bms-popup-surface-dark .message-media-control:last-child:not(:first-child) {
+  background-color: rgba(255, 255, 255, 0.10);
+}
```
krissedout: St does not implement functional :not() matching; it only matches pseudo-class names assigned to actors. This rule therefore never applies. Use supported child pseudo-classes -- for example style :last-child, then override :first-child:last-child for the single-control case, and fix the identical light/transparent rules.

src/styles/popup/dark.css
```diff
@@ -10,6 +10,61 @@
...
+
+.bms-popup-background-dark .message-media-control:hover,
+.bms-popup-surface-dark .message-media-control:hover,
+.bms-popup-background-dark .message-expand-button:hover,
+.bms-popup-surface-dark .message-expand-button:hover,
+.bms-popup-background-dark .message-list-section-collapse-button:hover,
+.bms-popup-surface-dark .message-list-section-collapse-button:hover,
```
krissedout: message-list-section-collapse-button is not a GNOME Shell style class in any supported 46–51 branch; the actor uses message-collapse-button. These selectors are dead, and the following StButton selector already covers the actual button anyway. Remove them or target the real class, including the duplicated light/transparent selectors.

krissedout (changes requested):

MeguMario:
Somehow the styling for when you touch (and hold) on the notification card just somehow not apply, I'm not sure whether this is css stuff or something we have to account for in popup blur infra ngl.

image

src/styles/popup/dark.css
```diff
@@ -10,6 +10,54 @@
...
 
+.bms-popup-background-dark .message-media-control,
+.bms-popup-surface-dark .message-media-control {
+  background-color: transparent;
+}
+
+.bms-popup-background-dark .message-media-control:last-child,
+.bms-popup-surface-dark .message-media-control:last-child {
+  background-color: transparent;
+}
```
krissedout: This replacement isn’t equivalent to the previous :not(:first-child) selector: every last control is now transparent, and the following sole-child rule redundantly sets it transparent again, so the intended multi-control highlight is gone. Restore the theme’s 0.10 background on :last-child, then let :first-child:last-child override it to transparent. The same correction is needed in light and transparent.

krissedout (changes requested):
> Somehow the styling for when you touch (and hold) on the notification card just somehow not apply, I'm not sure whether this is css stuff or something we have to account for in popup blur infra ngl. 

This shouldn’t need popup-blur infrastructure changes. Notifications are St.Buttons, and across GNOME 46–51 the press gesture applies :active on touch-begin and removes it on touch-end. The new .message:active selectors are therefore the correct hook.

For stacked notifications, this is a CSS cascade issue: the :second-in-stack and :lower-in-stack rules in the variant stylesheets load after base.css, have equal specificity, and override the active background. Add combined :second-in-stack:active / :lower-in-stack:active rules after them. For an unstacked card, inspect get_style_pseudo_class() while holding, if active is present, it’s CSS; if absent, it’s upstream input handling.

src/styles/popup/dark.css
MeguMario: I don't think `:last-child` is supposed to be highlighted outside of hovering on it, considering stock Adwaita doesn't have that either. THe one previously somehow managed to work unexpectedly so I missed that

image

image

krissedout: Then just remove both :last-child blocks entirely, since the base .message-media-control rule already makes them transparent. They’re redundant as written

MeguMario: here is with `:last-child` styled, which is not right

image

MeguMario: ok removed it

krissedout (approved):

MeguMario:
@krissedout ok so after some testing, I found out that the notification card does not supposed to do that and it's a behaviour of another extension (specifically Touch Up), so it's entirely my fault here. Sorry for the trouble 😭.

PR #977, opened 2026-09-01
Fix Dash-to-dock Panel mode not blurring on startup & Revert #963

### LINKED ISSUE #974: Issues with Dash to Dock after installing from source to get the popups settings
As the title says the blur effect stops working on the current version of Dash to Dock unless disabling and re-enabling the dash blurring or setting the artifact handling to performance then back to no artifacts.

https://github.com/user-attachments/assets/e6dda011-f706-4a69-8cff-d440452fd3e5

https://github.com/user-attachments/assets/3a3d6d84-bc3c-4c06-87bd-fd1987f05a4e

My system:
- CachyOS x86_64
- Linux 7.2.2-1-cachyos
- GNOME 50.4
- Mutter (Wayland)
- Dash to Dock version 105

Current Dash to Dock settings:

Image

Enabling "Use built-in theme" stops the blur effect from working what so ever.

### LINKED ISSUE #978: Performance optimization causing rendering artefacts (#963)
I've noticed black boxes appearing around UI elements when they rerender, after the changes from #963, which go away when I revert.
For example in the dock:
(sorry for the bad recording, the artefacts dont show up in a screen recording)

https://github.com/user-attachments/assets/fbf4d188-4af1-43d9-bf17-28f9589b6a41

The same effect also happens with the quick settings panel or the calender.

MeguMario:
- Fix #974 by tracking _slider signals and deferring dock update
- Revert #963 to fix #978
- Edit corner effects description in Static blur pipeline settings to now said that GNOME apps use a 15px radius instead of 12px (it's been 15px since GNOME 48)

### DIFF
--- a/src/components/dash_to_dock.js
+++ b/src/components/dash_to_dock.js
@@ -425,6 +425,15 @@ export const DashBlur = class DashBlur extends Signals.EventEmitter {
             _ => this.update_size()
         );
 
+        const slider = dash_container._slider;
+        if (slider) {
+            this.connections.connect(
+                slider,
+                ['notify::slide-x', 'notify::allocation'],
+                _ => this.update_size()
+            );
+        }
+
         const dash_background = dash._background ||
             dash.get_children().find(child => child.get_style_class_name?.()?.includes('dash-background')) ||
             dash;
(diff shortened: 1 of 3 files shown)


krissedout (approved):

MeguMario:
With this merged I think we should be ready for a new release @aunetx, now that #974 has been fixed.

PR #983, opened 2026-09-06
Popup blur stylesheet tweaks

MeguMario:
This introduce a small tweaks to popup blur stylesheet
- Style `modal-dialog` insensitive buttons
- Style `StEntry`
- Add missing styling for `screenshot-ui-panel` in light mode
image
image

There is this gnome shell bug related to OSK keep the key highlighted after hover or tap when in password prompt, we should report this to them as well (without extension it's there but faintly brighter)
Screenshot From 2026-09-06 18-15-35

### DIFF
--- a/src/styles/popup/base.css
+++ b/src/styles/popup/base.css
@@ -338,6 +338,18 @@
     background-color: rgba(255, 255, 255, 0.13) !important
 }
 
+.bms-popup-background-light .screenshot-ui-type-button,
+.bms-popup-background-light .screenshot-ui-show-pointer-button,
+.bms-popup-background-light .screenshot-ui-shot-cast-button {
+    color: #1a1a1e !important;
+}
+
+.bms-popup-background-light .screenshot-ui-type-button:insensitive,
+.bms-popup-background-light .screenshot-ui-show-pointer-button:insensitive,
+.bms-popup-background-light .screenshot-ui-shot-cast-button:insensitive {
+    color: rgba(26, 26, 30, 0.5) !important;
+}
+
 .bms-popup-background-transparent .screenshot-ui-type-button:hover,
 .bms-popup-background-dark .screenshot-ui-type-button:hover {
     background-color: rgba(255, 255, 255, 0.20) !important;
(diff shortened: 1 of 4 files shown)


src/styles/popup/transparent.css
```diff
@@ -66,6 +66,11 @@
     background-color: rgba(255, 255, 255, 0.18) !important;
 }
 
+.bms-popup-background-dark .modal-dialog .button:insensitive,
```
krissedout: the disabled-button selectors use "dark" instead of "transparent", so the intended styling never applies in Transparent mode.

src/styles/popup/dark.css
```diff
@@ -232,3 +237,15 @@
     background-color: rgba(255, 255, 255, 0.3);
     background-color: st-mix(-st-accent-fg-color, -st-accent-color, 30%);
 }
+
+.bms-popup-background-dark StEntry {
```
krissedout, line 241: entries get a translucent background without a matching foreground override. With a light Shell theme and forced Dark popups, text stays dark against the dark surface. Set readable normal/focused/disabled text colors too.

krissedout (changes requested):

src/styles/popup/dark.css
MeguMario: I changed it, but tbh using using force dark popups with light shell theme cause a lot of styling issue anyway.

src/styles/popup/base.css
```diff
@@ -406,6 +418,45 @@
     background-color: rgba(0, 0, 0, 0.16) !important;
 }
 
+.bms-popup-background-light .screenshot-ui-capture-button {
+    border: 4px #1a1a1e !important;
+}
+
+.bms-popup-background-light .screenshot-ui-capture-button .screenshot-ui-capture-button-circle {
+    background-color: #1a1a1e !important;
+}
```
krissedout, line 427: the new light-mode rules override the red capture button in video mode. I verified on GNOME 51: recording mode changes from red to the same dark color as screenshot mode. Preserve the red :cast styling, including hover/focus/active states.

krissedout (changes requested):

src/styles/popup/base.css
MeguMario: Should be red now

krissedout (approved):

PR #976, opened 2026-08-31
Replace Shell blur with a unified pipeline backend

krissedout:
This moves every blur implementation onto a shared rendering backend, replacing the separate component-specific paths and workarounds used throughout the extension.

What changed:

- Added a shared backend for static and dynamic blur surfaces
- Unified blur handling across popups, applications, app folders, the panel, dash, overview, lock screen, screenshots, and supported extensions
- Added dynamic backdrop capture
- Added Dual Kawase blur
- Reworked Liquid Glass and refraction rendering
- Added internal rounded-corner handling for all pipelines
- Added consistent enter and exit fading across effects
- Improved geometry, transforms, clipping, animation tracking, multi-monitor behavior, and resource cleanup
- Consolidated pipeline lifecycle and settings management
- Added migration of existing dynamic blur settings into pipelines
- Added Dhruva integration
- Updated preferences, schemas, styles, documentation, and translations

What was removed:

- Component-specific blur implementations
- Paint-signal capture and repaint hacks
- The third-party GNOME rounded blur dependency
- Its build script and setup documentation
- Native static and dynamic Gaussian blur backends
- The old GLSL Gaussian and Monte Carlo blur implementations
- Dummy pipeline handling
- The separate user-configurable corner effect
- Redundant cleanup and compatibility code

Testing notes:

The unusually large line count is primarily caused by regenerating the translation catalogs after the effects, preferences, and translatable strings changed.

The new backend has been tested on GNOME Shell 51 in a nested session, including the main popup, application, app-folder, animation, pipeline, and effect paths. It has not yet been tested exhaustively across every supported GNOME Shell version or with every third-party integration. Dash to Dock, Dash to Panel, Hide Top Bar, Window List, and other extension combinations still need broader testing. Given the size of the backend rewrite, additional bugs or integration-specific issues may remain.

Issue #931, opened 2026-07-19
[Feature] Dynamic Liquid Glass — real-time desktop refraction (Apple-quality, clone-based)


NB-Group:
### tl;dr

Real-time, **Apple-quality liquid glass** for the **panel, dash-to-dock and popups** — refracting the **live desktop** (wallpaper **and** windows, dynamic or static) through a clone-based backdrop + single-pass GLSL shader. Plus **three-way coexistence** with applications blur and dynamic wallpapers (e.g. waywallen).

🎬 **Demo:** 

https://github.com/user-attachments/assets/df189ed4-9bfe-4b9d-ad51-d7ea88b10039

> 📹 Handheld phone capture — OBS is bugged on this machine so I had to film the screen by hand. The actual on-screen effect is smooth; any flicker/tearing is a capture artifact, not the effect itself.

This is **not** BMS's existing "Liquid Glass" (`refraction.glsl`, which blurs a *static wallpaper snapshot*). That one can't follow windows or dynamic wallpapers. This is a new, clone-based pipeline that samples the **actual live stage** behind the widget, every frame.

### What it does

- **Live refraction**: the glass shows what's *actually behind it* right now — drag a window under the panel, switch workspaces, run a video wallpaper — it all refracts through the glass in real time.
- **Single-pass shader**: gaussian blur + refraction bands + top-bright/bottom-dark sheen + rounded-rect mask (ported from the apple-inspired-glass-effects-library reference).
- **Three-way coexistence** (the genuinely hard part): liquid-glass panel/dock/popup **+** applications blur **+** dynamic wallpaper — all at once, with windows showing through the glass.

### How (technical)

- New `conveniences/liquid_glass_pipeline.js` (clone-based see-through architecture): each glass widget clones `_backgroundGroup` + a shared `window_group` capture, then runs the `LiquidGlassEffect` shader on the widget.
- **Shared `window_group` capture** (one offscreen-cached clone of `window_group` in `uiGroup`): N glass widgets sample one cached texture, so the desktop is composited once per frame, not N times (avoids the mutter multi-clone pollution / window-shadow corruption you'd get from each widget cloning `window_group` directly).
- **applications blur coexistence**: applications keep BMS's native `Shell.BlurEffect(BACKGROUND)` inside `window_actor`. The whole-`window_group` capture includes that blur_actor → a **mild render loop**, deliberately accepted.

### Known limitations (this is the "experimental" part)

[...]

PR #985, opened 2026-09-07
Add new, more optimized glass effect

codasaii:
**This PR adds new glass effect as an alternative to the existing refraction effect**
- More optimized with less bloated glsl code without losing quality
- Visually match to Figma glass effect and looks very similar to MacOS 26
- Tuned to be more close as Figma as possible including refraction, depth, splay and other props

### Implementation
Most of this code logic taken from this article:
[Medium - Liquid Glass: iOS Effect Explanation](https://medium.com/@aghajari/liquid-glass-ios-effect-explanation-dabadd6414ae)

So basically this works by masking the actor using sdf, then remapping the distance from center this also adjust the depth, and add displacement using sphere projection formula to edge of the glass. 

For the splay effect, i did some experiment to find the best tune and it works by masking the displacement on the wide axis.

### Comparison

*GNOME Shell*
Screenshot 2026-09-07 at 10 39 35 AM

*MacOS 26.1*
Screenshot 2026-09-07 at 10 39 30 AM

*The dock looks so similar the mac 

*Figma Glass Effect*
Screenshot 2026-09-07 at 1 08 50 PM

*This kinda looks like jelly though*
image

### MacOS dock like pipeline settings
**Add these effects:**
1. Luminosity
   - Saturation: 1.05
   
2. Native gaussian blur
   - Radius: 0.8
   - Brightness: 0.95
   
3. Glass
   - Corner Radius: 24
   - Refraction: 100
   - Depth: 32
   - Splay: 0
   - Light Angle: 45º
   - Light Intensity: 10
   - Light ambient: 50
   - Light depth: 10
   - Light feather: 100

   

## One more issue
The clip padding under `get_clipped_bounding()` currently still being hard-coded because without it the glass looks out of bound or the bounding doesnt match with the actor bound. These hard-coded values works on my display but i don't know on other display and how to make this fit with the actor bounding

I assume that when actor be a texture, the size get offseted by 3 for example if the width is 100, the texture width is 103 so I tried to offset it by 3px but it doesnt work either and it still looks like it got cut off. Please help me with this...

<br><br>

*Hope this PR helps, thank you....*
Soon imma try to implement live liquid glass effect

### DIFF
--- a/src/effects/effect_groups.js
+++ b/src/effects/effect_groups.js
@@ -6,6 +6,7 @@ export function get_effects_groups(_ = _ => '') {
                 'native_static_gaussian_blur',
                 'gaussian_blur',
                 'monte_carlo_blur',
+                'glass',
             ],
         },
         texture_effects: {
--- a/src/effects/effects.js
+++ b/src/effects/effects.js
@@ -13,6 +13,8 @@ import { RgbToHslEffect } from './rgb_to_hsl.js';
 import { HslToRgbEffect } from './hsl_to_rgb.js';
 import { LuminosityEffect } from './luminosity.js';
 import { RefractionEffect } from './refraction.js';
+import { GlassEffect } from './glass.js';
+
 export { get_effects_groups } from './effect_groups.js';
 
 export function get_supported_effects(_ = () => "") {
@@ -493,6 +495,111 @@ export function get_supported_effects(_ = () => "") {
             description: _("Converts the image from HSLA colorspace to RGBA."),
             is_advanced: true,
             editable_params: {}
+        },
+
+        glass: {
+            class: GlassEffect,
+            name: _("Glass"),
+            description: _("A glass refraction effect that simulates a lens distorting the background."),
+            is_advanced: false,
+            editable_params: {
+                radius: {
+                    name: _("Corner radius"),
+                    description: _("The radius of the glass corners."),
+                    type: "integer",
+                    min: 5,
+                    max: 150,
+                    increment: 1,
+                },
+                refraction: {
+                    name: _("Refraction"),
+                    description: _("How strong the light refract along the edges."),
+                    type: "float",
+                    min: 0.,
+                    max: 100.,
+                    increment: 1.,
+                    big_increment: 10.,
+                    digits: 0.
+                },
+                depth: {
+                    name: _("Depth"),
+                    description: _("How thick your refraction appears from the edges."),
+                    type: "float",
+                    min: 0.,
+                    max: 50.,
+                    increment: 1.,
+                    big_increment: 10.,
+                    digits: 0.
+                },
+                dispersion: {
+                    name: _("Dispersion"),
+                    description: _("Chromatic aberration at the edges of your glass."),
+                    type: "float",
+                    min: 0.,
+                    max: 100.,
+                    increment: 1.,
+                    big_increment: 10.,
+                    digits: 0.
+                },
+                splay: {
+                    name: _("Splay"),
+                    description: _("The spread of how light bends around edges."),
+                    type: "float",
+                    min: 0.,
+                    max: 100.,
+                    increment: 1.,
+                    big_increment: 10.,
+                    digits: 0.
+                },
+                light_angle: {
+                    name: _("Light angle"),
+                    description: _("The direction the light hits the glass surface."),
+                    type: "integer",
+                    min: 0,
+                    max: 360,
+                    increment: 1,
+                },
+                light_intensity: {
+                    name: _("Light intensity"),
+                    description: _("How bright the rim light and shadow appear."),
+                    type: "float",
+                    min: 0.,
+                    max: 100.,
+                    increment: 1.,
+                    big_increment: 10.,
+                    digits: 0.
+                },
+                light_ambient: {
+                    name: _("Light ambient"),
+                    description: _("The base brightness applied evenly around the edges."),
+                    type: "float",
+                    min: 0.,
+                    max: 100.,
+                    increment: 1.,
+                    big_increment: 10.,
+                    digits: 0.
+                },
+                light_depth: {
+                    name: _("Light depth"),
+                    description: _("How far the rim light extends inward from the border."),
+                    type: "float",
+                    min: 0.,
+                    max: 10.,
+                    increment: 1.,
+                    big_increment: 2.,
+                    digits: 0.
+                },
+                light_feather: {
+                    name: _("Light feather"),
+                    description: _("How softly the rim light fades toward the center."),
+                    type: "float",
+                    min: 0.,
+                    max: 100.,
+                    increment: 1.,
+                    big_increment: 10.,
+                    digits: 0.
+                },
+            }
         },
 
         corner: {
--- a/src/effects/glass.glsl
+++ b/src/effects/glass.glsl
@@ -1,0 +1,152 @@
+uniform sampler2D tex;
+
+// Resolution
+uniform float width;
+uniform float height;
+
+// Clipping
+uniform float clip_x0;
+uniform float clip_y0;
+uniform float clip_width;
+uniform float clip_height;
+
+// Glass
+uniform float radius;
+uniform float refraction;
+uniform float depth;
+uniform float dispersion;
+uniform float splay;
+
+// Light
+uniform float light_angle;
+uniform float light_intensity;
+uniform float light_ambient;
+uniform float light_depth;
+uniform float light_feather;
+
+float sdf(vec2 p, vec2 b, float r) {
+    vec2 d = abs(p) - b + vec2(r);
+    return min(max(d.x, d.y), 0.0) + length(max(d, 0.0)) - r;   
+}
+
+float safe_sqrt(float val) {
+    return sqrt(max(val, 0.0));
+}
+
+float nonzero(float val){
+    return max(val, 000.1);
+}
+
+vec4 getTextureColorAt(vec2 coord) {
+    vec2 uv = coord / vec2(width, height);
+
+    if (uv.x < 2. / width)
+        uv.x = 2. / width;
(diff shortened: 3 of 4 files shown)


MuzahidGithub:
Liquid Glass is already being worked on in a separate pr with a better pipeline implementation. And you should not use any 26 series OS effects as reference, 26 was experiment and incomplete. Target iOS or MacOs 27

treebarkbr:
It's good.

But IMO the existing glass effect is already good enough. it was made to work with the upcoming refactor of the whole extension, and the refraction.glsl shader I authored for Blur my shell is a really close(ish) port of iOS 26's Liquid Glass. The original metal shader that refraction.glsl was ported from (Tweak.mm from Liquid (Gl)ass) is already really close to DnV1eX's LiquidGlassKit repo, minus the stuff stripped out to keep the necessities.

Let the maintainers handle this though

MeguMario:
Personally, I would like to see the current implementation get improve instead of having another seperated implementation (see https://github.com/aunetx/blur-my-shell/issues/931 for another one). I think this look really nice tbh, but we shouldn't need to have two effects which aim to replicate the liquid ass look

Also for the live liquid ass effect, please don't, that's the job of https://github.com/aunetx/blur-my-shell/pull/976

treebarkbr:
> Personally, I would like to see the current implementation get improve instead of having another seperated implementation (see #931 for another one). I think this look really nice tbh, but we shouldn't need to have two effects which aim to replicate the liquid ass look
> 
> Also for the live liquid ass effect, please don't, that's the job of #976

I'm currently looking forward to improve it but I'm somewhat stumped as of right now and I wanna see what other users say.

codasaii:
> It's good.
> 
> But IMO the existing glass effect is already good enough. it was made to work with the upcoming refactor of the whole extension, and the refraction.glsl shader I authored for Blur my shell is a really close(ish) port of iOS 26's Liquid Glass. The original metal shader that refraction.glsl was ported from (Tweak.mm from Liquid (Gl)ass) is already really close to DnV1eX's LiquidGlassKit repo, minus the stuff stripped out to keep the necessities.
> 
> Let the maintainers handle this though


Thank youu

I was thinking about not making this pr since someone had already worked on this, but i feel the implementation is more complex than mine. Bc i tryna have something like live liquid glass effect where the effect render real time behind it so the shader needs to be runs efficiently for better preformance. Maybe by making this pr, it would help them

tbh idk will this help or not, bc I was trying to optimize the current implementation but after i look into it further it uses different algorithm than mine so maybe having two shader would work here or maybe one for static one for real time shader idk...

this pr is just optional pr and yeah let's leave it to maintainers they know best

thank you for the feedback appreciate it 😁👍

MeguMario:
@codasaii 

> Bc i tryna have something like live liquid glass effect where the effect render real time behind it so the shader needs to be runs efficiently for better preformance. Maybe by making this pr, it would help them

> tbh idk will this help or not, bc I was trying to optimize the current implementation but after i look into it further it uses different algorithm than mine

I think the better option is to optimize the current one honestly. Don't worry about it having a different algorithm that much. It benefit both at the end of the day. 

If you like you can check out https://github.com/aunetx/blur-my-shell/pull/976, maybe your idea can help

> having two shader would work here or maybe one for static one for real time shader idk...

Since the goal is to unified both dynamic and static it's not a suitable solution tbh (it's still two different implementation need to be maintained at the end of the day)

I'll leave this open just in case you want to overhaul this. If you want it close feel free to ping me.

codasaii:
> @codasaii
> 
> > Bc i tryna have something like live liquid glass effect where the effect render real time behind it so the shader needs to be runs efficiently for better preformance. Maybe by making this pr, it would help them
> 
> > tbh idk will this help or not, bc I was trying to optimize the current implementation but after i look into it further it uses different algorithm than mine
> 
> I think the better option is to optimize the current one honestly. Don't worry about it having a different algorithm that much. It benefit both at the end of the day.
> 
> If you like you can check out #976, maybe your idea can help
> 
> > having two shader would work here or maybe one for static one for real time shader idk...
> 
> Since the goal is to unified both dynamic and static it's not a suitable solution tbh (it's still two different implementation need to be maintained at the end of the day)
> 
> I'll leave this open just in case you want to overhaul this. If you want it close feel free to ping me.

Hmm, is it okay to change most of refraction logic? (this also change the looks) but it can fix the weird diagonal line on the corners that prolly caused by heightSdf i had this same problem before when experimenting with liquid glass shader, mac also has this too so when corner radius is near 0, so i change the implementation to use another distortion formula

I might add or simplified some property and follow what figma already did with extra props so its easy to adjust this

We can also drop blur because we already native gaussian blur, I think tint color can be seperated to color so it reduces complexity

I was kinda afraid honestly to change the whole logic of current implementation bc its not longer Liquid Gl(ass) anymore with this

hmm what do you think? or maybe i need to wait till refactor is complete?

MeguMario:
@codasaii 

> but it can fix the weird diagonal line on the corners that prolly caused by heightSdf i had this same problem before when experimenting with liquid glass shader, mac also has this too so when corner radius is near 0, so i change the implementation to use another distortion formula

technically the corner bug is remedied with #976 by having the refraction rim constrained by the corner radius. tho for the use of another formula I would like to hear from @treebarkbr.

> We can also drop blur because we already native gaussian blur, I think tint color can be seperated to color so it reduces complexity

I think the tint can be relegated to the color effect as well, tho the blur in liquid gl(ass) is different compare to the one we have rn so personally I think it can stay. 

> I was kinda afraid honestly to change the whole logic of current implementation bc its not longer Liquid Gl(ass) anymore with this

> I might add or simplified some property and follow what figma already did with extra props so its easy to adjust this

Well, feel free to do it, it's what Pull Request is for (lol). 

> maybe i need to wait till refactor is complete?

Well you don't have to wait considering that one will take sometime to complete.

treebarkbr:
> > @codasaii
> > > Bc i tryna have something like live liquid glass effect where the effect render real time behind it so the shader needs to be runs efficiently for better preformance. Maybe by making this pr, it would help them
> > 
> > 
> > > tbh idk will this help or not, bc I was trying to optimize the current implementation but after i look into it further it uses different algorithm than mine
> > 
> > 
> > I think the better option is to optimize the current one honestly. Don't worry about it having a different algorithm that much. It benefit both at the end of the day.
> > If you like you can check out #976, maybe your idea can help
> > > having two shader would work here or maybe one for static one for real time shader idk...
> > 
> > 
> > Since the goal is to unified both dynamic and static it's not a suitable solution tbh (it's still two different implementation need to be maintained at the end of the day)
> > I'll leave this open just in case you want to overhaul this. If you want it close feel free to ping me.
> 
> Hmm, is it okay to change most of refraction logic? (this also change the looks) but it can fix the weird diagonal line on the corners that prolly caused by heightSdf i had this same problem before when experimenting with liquid glass shader, mac also has this too so when corner radius is near 0, so i change the implementation to use another distortion formula

#976 already solves this by constraining the bezel width with the corner radius, but the root cause I think is the normal computation in estimateAnalyticEdge. The refraction direction is computed with two different formulas, one in the corner region and one on the straight edges. In the corner it uses pow(cornerDelta, 3) which points diagonally and rotates smoothly, on the edges it's p - nearestCore which is hard axis aligned. Those two don't agree where they meet, so dir snaps at the junction and the sample position jumps by a small bit. That's the diagonal line that's popping up in the screenshot I shown of the artifact. It shows up once the refraction band extends past the corner region, which is the bezel > corner radius case (I haven't tested it vice versa yet). It's not from the height SDF, it's just the piecewise normal

That's just a guess though, I'll try talking to winaviation about this if I have the time to do so

> I might add or simplified some property and follow what figma already did with extra props so its easy to adjust this
> 
> We can also drop blur because we already native gaussian blur, I think tint color can be seperated to color so it reduces complexity

#976 already replaces my blur pass with the Dual Kawase blur

> I was kinda afraid honestly to change the whole logic of current implementation bc its not longer Liquid Gl(ass) anymore with this
> 
> hmm what do you think? or maybe i need to wait till refactor is complete?

MeguMario:
Small note is that as of version 0.1.1b of https://github.com/winaviation-tweaks/liquidass, the license was changed to CC BY-NC 4.0, so do keep in mind whenever anyone adapt the code from newer version of the shader code there since BMS is gpl-3.0 licensed

codasaii:
Yeah i seen #931 and it's shouldn't be two effects this will confuse user about which one should they use. And the implementation also has different results while the current use physical refraction that uses gl refract (maybe) mine uses straight formula based technique with tradeoff of accuracy but runs faster. 

I think I might actually send another pr #976 to propose to add switchable refraction type like dropdown, And is it ok to have two glsl file? afaik I don't think branching would be efficient in gl and I might retune the current implementation to looks more natural.

btw i got another effect called vapor for another pr, this effect its just like in the Avdan Showcase and thank you for the feedbacks....

MeguMario:
@codasaii If possible I prefer if it's in one glsl file. Aside from that feel free to contribute though.

treebarkbr:
> Small note is that as of version 0.1.1b of https://github.com/winaviation-tweaks/liquidass, the license was changed to CC BY-NC 4.0, so do keep in mind whenever anyone adapt the code from newer version of the shader code there since BMS is gpl-3.0 licensed

Noted. I'll be asking winaviation if it's okay to port the new shader over

Edit: 0.1.1b rewrote everything so porting this may take a while lol... and the formula is now different it throws out the raytracing bits
I'll start attempting to port it right now and I'll open a PR if I get permission

Edit (10:35PM): I got permission.

krissedout:
If I understood the discussion correctly, the plan is to bring these ideas into the existing Liquid Glass effect rather than maintain a separate effect. In that case, I think we can close this one. Thanks for experimenting with this and sharing the implementation, and any future contributions are definitely welcome!

PR #987, opened 2026-09-08
Update Liquid Glass effect with Liquid (Gl)ass 0.1.1b shader

treebarkbr:
## Summary

Updates the Liquid Glass refraction effect with the shader features from winaviation's Liquid (Gl)ass tweak 0.1.1b, building on the 0.1.0b refraction work and the unified rendering backend #976. The changes are ported from `LiquidAssBackboardd/Tweak.mm` (Metal) to GLSL for Blur my Shell.

## How it works

Builds on the 0.1.0b refraction work and updates the shader to 0.1.1b:

- Convex-circle refraction: #976 swapped 0.1.0b's squircles for convex-circle rounding to match GNOME's standard rounded corners
- Chromatic dispersion accumulation (3 + 4 taps split along the fresnel glare axis)
- Quartz edge profile (`1 - sqrt(t*(2-t))`) and highlight, keeping the edge falloff and glare band tight to the rounded-corner rim
- A fresnel glare angle setting (-180..180 deg, default +45 deg, 0 = light from above, positive clockwise) that rotates the glare and dispersion axis to match the original shader orientation
- A fresnel glare width setting (default 1.0)
- An optional specular glare that shows the light direction across the glass face, off by default (addition beyond the 0.1.1b shader) that scales the highlight ring thickness up or down
- Bezel width range widened to 500 and glass thickness to 20 for heavier glass looks
- The un-premultiplied backdrop sampling, tint and shadow are unchanged

## Known issues

- Not ported from 0.1.1b: glyph mask refraction (iOS clock and icons), and the coversheet mode, which refracts the lock screen around the edges when you swipe up to unlock. Neither maps to a surface Blur my Shell has

## Testing

- Tested manually on GNOME 50.3, Wayland
- `make build` succeeds

## License

`refraction.glsl` is a port of winaviation's Liquid (Gl)ass 0.1.1b. winaviation confirmed in this PR (2026-09-11) that this ported shader file is re-licensed under **GPLv3** for blur-my-shell, with the same rights extending to anyone who reuses it from this repository. The original upstream project remains **CC BY-NC 4.0**. Attribution to the original work is retained in the file header.

### DIFF
--- a/src/effects/effects.js
+++ b/src/effects/effects.js
@@ -330,7 +330,7 @@ export function get_supported_effects(_ = () => "") {
                 },
                 fresnel_angle: {
                     name: _("Fresnel glare angle"),
-                    description: _("Light direction of the fresnel glare, in degrees (0 is to the right, 90 is down)."),
+                    description: _("Light direction of the fresnel glare, in degrees (0 is from above, positive rotates clockwise)."),
                     type: "float",
                     min: -180.,
                     max: 180.,
--- a/src/effects/effects.js
+++ b/src/effects/effects.js
@@ -350,7 +350,7 @@ export function get_supported_effects(_ = () => "") {
                 },
                 specular_strength: {
                     name: _("Specular glare strength"),
-                    description: _("Strength of the specular glare showing the light direction (toggle in preferences)."),
+                    description: _("Strength of the specular glare showing the light direction (0 disables it)."),
                     type: "float",
                     min: 0.,
                     max: 1.,
--- a/src/effects/effects.js
+++ b/src/effects/effects.js
@@ -348,6 +348,16 @@ export function get_supported_effects(_ = () => "") {
                     big_increment: 0.5,
                     digits: 2
                 },
+                specular_strength: {
+                    name: _("Specular glare strength"),
+                    description: _("Strength of the specular glare showing the light direction (toggle in preferences)."),
+                    type: "float",
+                    min: 0.,
+                    max: 1.,
+                    increment: 0.01,
+                    big_increment: 0.1,
+                    digits: 2
+                },
                 tint: {
                     name: _("Tint strength"),
                     description: _("Amount of subtle milky glass tint over the blurred texture."),
--- a/src/effects/effects.js
+++ b/src/effects/effects.js
@@ -293,7 +293,7 @@ export function get_supported_effects(_ = () => "") {
                     description: _("How far the liquid-glass lens reaches inward from the edge."),
                     type: "float",
                     min: 1.,
-                    max: 100.,
+                    max: 500.,
                     increment: 1.,
                     big_increment: 10.,
                     digits: 0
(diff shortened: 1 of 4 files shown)


MeguMario:
Now this is is diffcult, I'm not really well versed with licensing stuff, but from what I know, fundamentally CC BY-NC 4.0 is incompatible with GPL-3.0. So this is hard for me to merge this. I'll let @aunetx has the last say in this though. But in the worst case we will have to stick with 0.1.0b imo. 

Anyone well versed in licensing can help me out in this though. Not really confident in this 😭

krissedout:
Thanks for working on this!

For the licensing, we’d need winaviation to explicitly allow this shader and its GLSL port to be distributed under GPL-3.0, with the same rights extending to downstream users. Them saying we can distribute it under the existing license isn’t enough to resolve this, since the non-commercial restriction still applies and conflicts with GPL-3.0.

There’s also a corner artifact from allowing the refraction rim to extend five pixels beyond the corner radius. I reproduced a stepped colour artifact with square corners, and restoring the existing radius limit removes that wedge. Could we keep that constraint for now, or adjust the edge calculation so extending the rim doesn’t bring the seam back?

treebarkbr:
> Thanks for working on this!
> 
> For the licensing, we’d need winaviation to explicitly allow this shader and its GLSL port to be distributed under GPL-3.0, with the same rights extending to downstream users. Them saying we can distribute it under the existing license isn’t enough to resolve this, since the non-commercial restriction still applies and conflicts with GPL-3.0.
> 
> There’s also a corner artifact from allowing the refraction rim to extend five pixels beyond the corner radius. I reproduced a stepped colour artifact with square corners, and restoring the existing radius limit removes that wedge. Could we keep that constraint for now, or adjust the edge calculation so extending the rim doesn’t bring the seam back?

Thanks a lot for the reply

I already let winaviation know about this as of now and I'll follow up if he ever replies, I attached the PR to the email as well.

Anyways, addressed the feedback and reverted the looser clamp

Cheers

winaviation:
hi, the maintainers can merge. no problem. i switched to CC BY-NC 4.0 mostly to stop ppl from selling my shit lol

krissedout:
@winaviation Thanks for confirming! Could we just clarify the shader header before merging? It currently says GPLv3 applies, but then says non-commercial restrictions still apply outside Blur my Shell. If the permission covers this version under GPLv3, those rights need to carry over when someone reuses it elsewhere too. The original upstream project can keep its existing license; we just need the header to clearly distinguish that from the permission granted here.

winaviation:
> @winaviation Thanks for confirming! Could we just clarify the shader header before merging? It currently says GPLv3 applies, but then says non-commercial restrictions still apply outside Blur my Shell. If the permission covers this version under GPLv3, those rights need to carry over when someone reuses it elsewhere too. The original upstream project can keep its existing license; we just need the header to clearly distinguish that from the permission granted here.

ykw i'm fine with relicensing this specific ported shader file under GPLv3 for blur my shell (and anyone who reuses it from your repo). my original upstream will stay CC BY-NC 4.0.

feel free to update the header to standard GPLv3 like so, or wtv
```
// GLSL port of liquidass (LiquidAssBackboardd/Tweak.mm kShaderSrc)
// original shader by winaviation (https://github.com/winaviation-tweaks/liquidass)
// ported and re-licensed under GPLv3 for Blur my Shell with permission from the author.
```

krissedout:
The licensing clarification and updated header resolve my concern, thanks! 

@treebarkbr I found one remaining issue with the thickness control: the new profile caps glassThickness at refractionBand. With bezel width 22 and corner radius 26, that means increasing thickness above roughly 2.15 does nothing, I tested 2.4 versus 20 and got identical output. Could we keep thickness responsive independently of the rim width while preserving the corner constraint?

treebarkbr:
> The licensing clarification and updated header resolve my concern, thanks!
> 
> @treebarkbr I found one remaining issue with the thickness control: the new profile caps glassThickness at refractionBand. With bezel width 22 and corner radius 26, that means increasing thickness above roughly 2.15 does nothing, I tested 2.4 versus 20 and got identical output. Could we keep thickness responsive independently of the rim width while preserving the corner constraint?

Noted, I'll update it in a bit as I'm also wrestling with the shader as I'm also trying to add specular highlights.

treebarkbr:
Done.

krissedout:
Thanks, thickness responds now! Removing the cap exposes one issue on fully circular surfaces: their displacement uses edgeOpacity instead of edgeBand, so it abruptly drops to zero at refractionBand and creates an inner-ring seam. I reproduced this with a 240×240 surface, radius 120, bezel 22 and thickness 20. Using edgeBand for the circular displacement too removed the jump while keeping thickness responsive.

treebarkbr:
> Thanks, thickness responds now! Removing the cap exposes one issue on fully circular surfaces: their displacement uses edgeOpacity instead of edgeBand, so it abruptly drops to zero at refractionBand and creates an inner-ring seam. I reproduced this with a 240×240 surface, radius 120, bezel 22 and thickness 20. Using edgeBand for the circular displacement too removed the jump while keeping thickness responsive.

Fixed (hopefully), try testing it now

krissedout (approved):
LGTM now! Do you have anything else to finish before we merge?

treebarkbr:
Everything seems fine, ready to merge.

PR #988, opened 2026-09-09
Add Dynamic Vapor Effect 

codasaii:
**This pr adds new Vapor Wave effect, make you desktop feels more dynamic and alive**
- Realtime flowing animation (animation can be disabled)
- Unified animation looper for all vapor wave shaders (for efficiency)
- Visually looks close to original Avdan Vapor UX (for now)

This effect is different from Apple Liquid Glass, while apple liquid glass add refraction to edges of box this effect add wavy displacement for the whole surface and this mimic how liquid vapor behave. All ideas taken from one of the amazing Avdan 's Video:

[Reference](https://www.youtube.com/watch?v=8-yLnMBdjMI)

## Implementation (How this works)
This works by offseting the fragcoord randomly using fbm function, when texture is sampled from offseted and random coordinate, it will create something like wavy displacement and it can be tune to looks more liquidy.

The noise and fbm function is taken from different sources I use the most efficient noise generator as I can find and fbm is taken from the book of shaders article that explain how to generate random wavy function

## Showcase

Screenshot 2026-09-08 at 3 56 39 PM

https://github.com/user-attachments/assets/3ff2b8fc-fd72-4d8d-bee4-122f45de69d1

*Sorry if video looks pixalated, github has 10mb upload limit so I had to compress it via Whatsapp

This looks kinda cool tho
image
<br>
*I Can't find shell theme that looks like avdan os and it might looks different from reference video, because there's no border and rounded corner that match with the reference

## Worth to mention
- Performance: The animation might be heavy on low end device, I tried to fix it by using unified ticker for the timer, so far it runs well in my VM on my M2 device (6gb, 2cores, virt-gpu configuration), I believe the new refactor from #976 will Improve performance in the rendering side
- Visual: I feels this can be tune to be more similiar to the ref video, I need your opinion and feedback of what missing from this shader and some recommendations would help me to improve this
- Code: WaveTicker currently still in the one file with wave.js, this could be extracted and refactored somewhere but im not sure because im still waiting for #976 to be finished

* Sorry if my writing is bad 😅


MuzahidGithub:
instead of adding this to blur my shell ( it supposed to do blur only ) you can create a separate extension with blur my shell as a requirement, then dynamically inject your modifications on each users BMS stylesheets. Maybe you can name it "Blur my shell effects" or "Shell Effects", this way you can have many different effects not only waves and use them on demand

MeguMario:
> instead of adding this to blur my shell ( it supposed to do blur only )

@MuzahidGithub I don't think so, considering the pipeline system allow for extra effects (that's not blur) to be use on top of blur, so this fit in just fine

src/effects/wave.glsl
```diff
@@ -0,0 +1,116 @@
...
+    float frequency = 0.;
+
+    // Loop of octaves
+    for (int i = 0; i < OCTAVES; i++) {
+        value += amplitude * noise(st);
+        st *= 2.;
+        amplitude *= .5;
+    }
+    return value;
+}
```
krissedout, line 69: Could you clarify the licenses for the copied noise and fBM implementations? The Book of Shaders currently has an all-rights-reserved license that prohibits redistribution, and the gist author also mentions licensing problems with that collection. We need clearly GPL-compatible implementations with the required attribution/license notices, or explicit permission from the relevant copyright holders.

src/effects/wave.js
```diff
@@ -0,0 +1,405 @@
...
+                this.update_animation_state();
+            }
+        }
+
+        update_animation_state() {
+            if (!this._update_time_ref) {
+                this._update_time_ref = this.update_time.bind(this)
+            }
+            
+            const active = this._can_animation && this._use_animation && this._speed_factor !== 0 && !!this.get_actor();
```
krissedout, line 319: This keeps the ticker running whenever the effect is attached, even when its actor or an ancestor is hidden. I tested this on GNOME 51: a hidden parent still produced 30 callbacks in 500 ms, and an actor with zero opacity still caused 31 redraws. Could you suspend ticker participation while the surface isn’t visible and resume it when it becomes visible again?

src/effects/wave.js
```diff
@@ -0,0 +1,405 @@
...
+                this._timeline.disconnect(this._connectionId);
+                
+                this._connectionId = undefined;
+            }
+
+            this.tasks.clear();
+        }
+}
+
+const WaveEffectClass = utils.IS_IN_PREFERENCES ? null : class WaveEffect extends Clutter.ShaderEffect {
```
krissedout, line 180: This needs to use the shared shader base from #976 instead of extending Clutter.ShaderEffect directly. I tested the combined branches: they build, but at runtime uniform upload fails with “effect.set_surface_uniform is not a function” and the wave effect isn’t rendered.

src/effects/wave.js
```diff
@@ -0,0 +1,405 @@
...
+            this.update_animation_state(enabled);
+
+            super.vfunc_set_enabled(enabled);           
+        }
+
+        vfunc_dispose() {
+            this._can_animation = false;
+            this.update_animation_state();
+
+            super.vfunc_dispose();
```
krissedout, line 399: Please remove the JavaScript dispose override and handle cleanup explicitly through the effect’s detach/lifecycle path. In both GNOME 51 test sessions, GJS blocked this callback during garbage collection with “Attempting to run a JS callback during garbage collection.” Disposal can happen when JavaScript execution is unsafe, so we can’t rely on this for cleanup.

krissedout (changes requested):

codasaii:
Thanks for the feedback @krissedout 
> The Book of Shaders currently has an all-rights-reserved license that prohibits redistribution

I don't know that The Book of Shaders has license that is not compatible with GNU GPL, I don't really know much about license 😅

I found several Implementation of fbm:
- This is one is from godot [Godot - Fractal Brownian Motion (fBM)](https://godotshaders.com/snippet/fractal-brownian-motion-fbm/) and as my research, Godot uses MIT license that I found it compatible with GNU GPL, but I can't find what license this shader under, because from [License - Godot Shaders](https://godotshaders.com/license/) the shader could use [CC0 (public domain)](https://creativecommons.org/publicdomain/zero/1.0/) which i don't kno wether this compatible or not
- This one is from shadertoy [FBM Scaling](https://www.shadertoy.com/view/WsV3zz) it uses MIT License on the header file, this the second alternative I might use

Btw I want to use fbm fn from godot but im not sure about the license, and what format should I use when putting it on top of code (idk is this necessary)? maybe like this?
`​// Based on <SymbolName> from <project_name> - (<License>) Copyright (c) <year> <holder>`
`// <link>`

codasaii:
> a hidden parent still produced 30 callbacks in 500 ms, and an actor with zero opacity still caused 31 redraws.

Yeah I forgot for a case when actor is not rendered. Is using actor.is_mapped and notify::map correct here? Afaik it's wether actor is being rendered or not

> This needs to use the shared shader base from https://github.com/aunetx/blur-my-shell/pull/976 instead of extending Clutter.ShaderEffect directly

I forgot to update this, ok imma change this to utils.ShaderEffect

> Please remove the JavaScript dispose override and handle cleanup explicitly through the effect’s detach/lifecycle path.

Yeah I shouldn't use this, I don't know when effect being detached from actor will it call `vfunc_set_actor` and `vfunc_set_enabled` or not, or maybe by using two of these vfunc is enough to avoid ticker leak??

Btw the effect needs more tuning, because in my test the wave looks rough instead of smooth liquid vapor in bigger frame, here's the current tuning
image

The white point on center is for contrast check wether this survive on dark and white bg or not

MeguMario:
@codasaii 

> because from License - Godot Shaders the shader could use CC0 (public domain) which i don't kno wether this compatible or not

You can use CC0 or MIT code freely in GPL-3.0 project like this, it's only the other way around that it's incompatible

krissedout:
MIT and CC0 implementations are fine to use here, but we need to verify the license of the specific snippet, not Godot’s engine license. The Godot Shaders page you linked doesn’t clearly identify its license, so I wouldn’t assume it’s CC0. If the ShaderToy implementation explicitly has an MIT header, preserve its copyright notice and full license text alongside the copied code; the short attribution comment alone isn’t enough.

For visibility, use actor.mapped or actor.is_mapped(), and notify::mapped rather than notify::map. Mapping handles hidden ancestors, but doesn’t cover zero opacity, so also account for effective opacity and react to relevant ancestor opacity changes. The shared ticker should stop when no effects need animation.

Detaching does call vfunc_set_actor(null), including when the actor is destroyed. Remove the ticker callback and disconnect the old actor’s signals there. Use vfunc_set_enabled for disabling/re-enabling separately--detaching doesn’t necessarily disable the effect. No dispose override is needed.

For the roughness on larger surfaces, the current noise coordinates are pixel-based and displacement grows with distance from the center. I’d start by making the noise scale consistent across surface sizes and bounding the displacement independently of surface size. Increasing octaves would add detail and cost, not necessarily make it smoother.

codasaii:
image

OK I think this feels smoother now, on bigger surface it's not really subtle so its not really distracting, but for smaller area the effect is more justified for better micro interaction. I use simplex noise that is more natural for this replacing the old hash based noise function.

codasaii:
Hmm i think i found compatibility bug in corner.glsl and refraction.glsl where `fwidth` is not recognize by cogl compiler and gives me this error: `0:97(29): error: no function with name 'fwidth'`

after some research it says that the driver is lack of derivative support or it's not properly supported, so maybe cogl doesn't support derivative functions or it is still using old gl. I tested using `dFdx` and it gives me the exact same error. Hmm I think on some devices derivative functions it's not supported by cogl, I tested on vm btw.

krissedout:
Thanks for the update! I retested on GNOME 51, and the shared shader integration and cleanup work now. Hidden parents, zero opacity on the actor itself, disabling the effect, and detaching it all stop the animation correctly.

There’s still one visibility case missing: a parent with opacity 0 leaves the ticker running. I measured 30 callbacks in 500 ms. The code only checks and watches the actor’s own opacity, so please account for effective opacity and ancestor opacity changes, including when the actor is reparented.

The replacement noise source is fine to use under MIT, but please include its full copyright and permission notice with the shader. The current attribution and links aren’t enough; putting the upstream notice in the shader header would work.

About the fwidth/dFdx error: thanks for catching that. Those calls are in #976’s existing shaders, so we should investigate it separately from this effect. They compile on my GNOME 51/NVIDIA setup, so it doesn’t appear to be a universal Cogl limitation. Could you share your VM’s GNOME/Mutter versions, renderer and GL/GLES version, and the full shader compilation log? That should help narrow down whether it’s a shader-language configuration issue or a driver limitation.

Proman4713:
I like this, but I don't think Avdan-style UI will appeal to a larger audience; some may find it nauseating (or it could make lower-end systems struggle). In my opinion, UX should be a major consideration alongside UI. You haven't modified the gschema XML either, so it seems that you don't have a toggle for this effect.

MeguMario:
@Proman4713 it is not a toggle but instead an effect you can add in BMS to use alongside Dual Kawase Blur and Liquid Glass in pipeline settings, so it's always an option you can choose

Proman4713:
@MeguMario makes sense, I missed that so it's good to know

codasaii:
> effective opacity and ancestor opacity changes, including when the actor is reparented.

Sure thing, btw is using notify::opacity correct here or there's notify::paint-opacity signals? and i'm also not sure which to use between parent-set and notify::parent signal



> full copyright and permission notice

I'm not sure about this, is that the full MIT License text from LICENSE or some small attribution from this file https://github.com/stegu/webgl-noise/blob/master/src/noise2D.glsl? maybe for project with different license than MIT, the whole text of LICENSE must be included?

codasaii:
@krissedout 

> fwidth/dFdx error:
## Vm Info
VM Device: Macbook Air M2
VM App: UTM
Arch: ARM64
System: QEMU 10.0 ARM VM 
Cores: 4
Ram: 6144 MB
gpu: virtio-gpu-gl-pci



## Gnome Version:
```bash
❯ gnome-shell --version
mutter --version
GNOME Shell 48.8
mutter 48.8
```

Also tested on GNOME 50, doesn't work either

## GL Info
```bash
❯ glxinfo -B | grep -E "OpenGL (renderer|version|vendor)"
OpenGL vendor string: Mesa
OpenGL renderer string: virgl (ANGLE (Apple, Apple M2, OpenGL 4.1 Metal - 90.5))
OpenGL version string: 2.1 Mesa 25.1.9
```
Yep, I think because of OpenGL 2.1 does not support derivative function, so cogl has no way to call this function nor cogl provide fallback for derivative functions

## Full Compilation log
```bash
Sep 14 07:25:14 fedora gnome-shell[3851]: Shader compilation failed:
                                          0:44(42): warning: `cogl_texel0' used uninitialized
                                          0:97(29): error: no function with name 'fwidth'
                                          0:97(25): error: no matching function for call to `max(error, float)'; candidates are:
                                          0:97(25): error:    float max(float, float)
                                          0:97(25): error:    vec2 max(vec2, float)
                                          0:97(25): error:    vec3 max(vec3, float)
                                          0:97(25): error:    vec4 max(vec4, float)
                                          0:97(25): error:    vec2 max(vec2, vec2)
                                          0:97(25): error:    vec3 max(vec3, vec3)
                                          0:97(25): error:    vec4 max(vec4, vec4)
                                          0:97(25): error:    int max(int, int)
                                          0:97(25): error:    ivec2 max(ivec2, int)
                                          0:97(25): error:    ivec3 max(ivec3, int)
                                          0:97(25): error:    ivec4 max(ivec4, int)
                                          0:97(25): error:    ivec2 max(ivec2, ivec2)
                                          0:97(25): error:    ivec3 max(ivec3, ivec3)
                                          0:97(25): error:    ivec4 max(ivec4, ivec4)
                                          0:99(3): warning: `antialiasWidth' used uninitialized
                                          0:100(2): warning: `antialiasWidth' used uninitialized
Sep 14 07:25:14 fedora gnome-shell[3851]: Failed to link GLSL program:
```

My solution is I think we can use safer function to avoid this problem in some devices. I haven't found the alternative yet, I think smoothstep would works btw what this `max(fwidth(distance), 0.0001);` does in corner.glsl?

MeguMario:
Personally, I think extending support to OpenGL 2.1 is a little bit too much, with the exception of Apple computers, computers that only support to maximum OpenGL 2.1 should better off not using BMS at all (2.1 is 20 years old now). Recent computers (from around  2016 to now) should support 4.6 just fine (with ofc, Apple as the exception).

My opengl version output for context, not the best computer but modern enough to have decent graphics api support
```
❯ glxinfo -B | grep -E "OpenGL (renderer|version|vendor)"
OpenGL vendor string: Intel
OpenGL renderer string: Mesa Intel(R) HD Graphics 620 (KBL GT2)
OpenGL version string: 4.6 (Compatibility Profile) Mesa 26.3.0-devel (git-0b1b8798d8)
```

MuzahidGithub:
@codasaii no offense just asking. are you using ai for all these? I have a feeling

krissedout:
@codasaii 

Yep, notify::opacity is the one you want, but watch the parents too and check get_paint_opacity() for the combined value. Use parent-set to reconnect when the hierarchy changes, including when an ancestor gets moved. There isn’t a notify::paint-opacity signal. And yes I meant the full copyright and MIT notice from LICENSE, not just the short attribution in the shader.

Also, I tracked down the shader error and pushed a fix in #976! Our shaders were assuming derivatives were available on Cogl’s GLES path. Simply enabling the extension wasn’t enough either, because Mesa rejects the declaration where Cogl inserts it. That path now calculates the antialiasing width directly, while desktop GL keeps the existing behaviour.

Both shaders compile on Mesa and NVIDIA, and I tested them in a GNOME 51 GLES session too. Could you retest with the latest #976? This was on our side, not something your Vapor effect needed to fix.

codasaii:
> @codasaii no offense just asking. are you using ai for all these? I have a feeling

I use ai, but not for all. I google it to find suitable shader code for this like from shaderToys, the book of shader, godot shader, or searching on github to find opensource function that can be use with this. The shader also requires tuning to make it looks good.

AI does really helps when dealing with terminology of what this fn or var should be called, catching bugs, and helping me to understand license. I would check my code again if ai catch bug, then fix it. Honestly ai code is more complicated than as it should so i dont use ai code for this. I think this use of ai is okay maybee... 😅

codasaii:
> @codasaii
> 
> Yep, notify::opacity is the one you want, but watch the parents too and check get_paint_opacity() for the combined value. Use parent-set to reconnect when the hierarchy changes, including when an ancestor gets moved. There isn’t a notify::paint-opacity signal. And yes I meant the full copyright and MIT notice from LICENSE, not just the short attribution in the shader.
> 
> Also, I tracked down the shader error and pushed a fix in #976! Our shaders were assuming derivatives were available on Cogl’s GLES path. Simply enabling the extension wasn’t enough either, because Mesa rejects the declaration where Cogl inserts it. That path now calculates the antialiasing width directly, while desktop GL keeps the existing behaviour.
> 
> Both shaders compile on Mesa and NVIDIA, and I tested them in a GNOME 51 GLES session too. Could you retest with the latest #976? This was on our side, not something your Vapor effect needed to fix.

Yess it worked now, the compatibility issue is fixed on my vm, thank youu... 
btw is the animation state works well now? I feel it's really smooth now even with dynamic mode in my vm. I also have a plan to add feature so user select to only blur libadwaita window only or select window type to blur like dialog, menu, etc.

krissedout:
Glad the VM compatibility fix worked! The immediate-parent opacity case is fixed now, and the license notice is included too. There’s still one animation case left: only the direct parent is watched, so a grandparent becoming transparent, or an ancestor moving under a transparent container, leaves the ticker running. Please watch opacity and parent changes along the whole ancestor chain, reconnecting when the hierarchy changes.

The window-type filtering idea sounds worth discussing separately, but I’d keep this PR focused on the effect.

PR #991, opened 2026-09-11
Add Copyous extension blur support

tobih7:
Adds optional blur support for the Copyous clipboard manager extension.
Since Copyous creates a popup, the existing popup code is used.
The toggle is in Other, below Coverflow.

Also fixes modal blur placement so the edit dialog blurs the clipboard popup behind it. (Thanks to Codex, because I had absolutely no idea how to fix that.)

Works with GNOME Shell 50.4 and Copyous 2.0.1. 
I tested dynamic/static blur, toggle restoration, and ran the tests.

### DIFF
--- a/resources/ui/other.ui
+++ b/resources/ui/other.ui
@@ -115,6 +115,18 @@
 
     <child>
       <object class="AdwPreferencesGroup">
+        <property name="title" translatable="yes">Copyous extension blur</property>
+        <property name="description" translatable="yes">Blur the Copyous clipboard popup. Uses the blur settings and dialog corner radius from the Popups page. Popup blur must be enabled.</property>
+        <property name="header-suffix">
+          <object class="GtkSwitch" id="blur_copyous">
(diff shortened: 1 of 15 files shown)


MuzahidGithub:
it should automatically get blurred if it uses the same popup surface. i feel like it should be done in copyous side instead of adding dedicated support on bms codebase, many other extensions could follow and it will get messy fast. Reuse existing common surfaces

MeguMario:
Since this is just an extension of Popup blur, I feel like you should not put a switch in Other (those that is in Other are special cases that outside of Popup blur) and consolidate that into Other popup type instead

Since its popup you should just let popup blur handle styling instead of doing it yourself as well

tobih7:
Yes, that makes sense. I'll loook into Copyous itself, it should properly identify it's popup as `modal-dialog` / `popup-menu-content`.

But there still is the modal dialog stacking bug. See #992.

PR #992, opened 2026-09-12
Fix blur stacking for modal dialogs

tobih7:
While working on PR #991 I found a general bug in the handling of stacked modal dialogs.

Currently blurs for modal dialogs are placed below the entire `modalDialogGroup`. This works when there is only one dialog, but when another modal dialog is opened on top of an existing one, its blur is still below both dialogs and therefore cannot blur the dialog behind it.

The fix treats `modalDialogGroup` as the parent for these blur actors and places each blur directly below its own dialog. It also starts the parent lookup from the actual blur target instead of the root of the scan. So multiple dialogs found below the same root get the correct indiviudal stacking position.

Since modal dialogs use `Clutter.BinLayout`, simply moving the blur into `modalDialogGroup` is not enough. The layout manager can override the blur actor's geometry. The added allocation constraint makes sure the blur keeps the exact position and size calculated for the dialog.

This is what I mean:

### Current master branch
master

### Fixed version
fixed

This test can be reproduced with [this test extension](https://github.com/aunetx/blur-my-shell/tree/994e6daa40014da56c33e74ae9c1940211cb27a4/bms-modal-blur-reproducer).

### DIFF
--- a/src/components/popup/surface_placement.js
+++ b/src/components/popup/surface_placement.js
@@ -2,6 +2,7 @@ import Graphene from 'gi://Graphene';
 import * as Main from 'resource:///org/gnome/shell/ui/main.js';
 
 import { PopupBlurSurfaceGeometry, transform_to_actor_space } from './surface_geometry.js';
+import { PopupBlurAllocation } from './surface_allocation.js';
 
 const MIN_SURFACE_DIMENSION = 2;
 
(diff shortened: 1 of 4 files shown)


MeguMario:
This was fixed in #976 if I remember

Edit: Ah no seems like it's not

tobih7:
#976 still has the same issue.
image

MeguMario:
@tobih7 Sorry, I mistook it for another similar bug that we used to have. Let me go through this real quick.

krissedout:
Could you please retarget this to refactor/unified-blur-backend and rebase it onto that branch?

tobih7:
done!

src/components/popup/surface_placement.js
```diff
@@ -295,6 +296,16 @@ export const PopupBlurSurfacePlacement = class PopupBlurSurfacePlacement {
...
                 return false;
 
+            if (this.surface.parent === Main.layoutManager.modalDialogGroup
+                && (!this.allocation_constraint || this.width !== local.width || this.height !== local.height)) {
+                if (!this.allocation_constraint) {
+                    this.allocation_constraint = new PopupBlurAllocation();
+                    this.surface.blur_actor.add_constraint(this.allocation_constraint);
+                }
+                // The transform positions the blur; constrain its local bounds only.
+                this.allocation_constraint.set_geometry(0, 0, local.width, local.height);
```
krissedout, line 306: Thanks, the rebase looks good and dynamic blur keeps the correct positioning. There’s still an issue with static blur though: its background container also moves into modalDialogGroup, but this constraint only handles dynamic blur. I reproduced the static container being allocated at the screen centre with zero size, which offsets the wallpaper blur. Could you handle the static container’s allocation too, preserving its origin and the background’s existing monitor-relative positioning and clip?

krissedout (changes requested):

src/components/popup/surface_placement.js
tobih7: Yes, I forgot to test static blur. It's working now.

krissedout (approved):

PR #996, opened 2026-09-13
Fix stylesheet priority

tobih7:
## The Issue

* Currently all CSS files are imported through `@import` in `stylesheet.css`.
* GNOME Shell gives stylesheets loaded directly by extensions a higher cascade priority, but stylesheets loaded through `@import` do not get that priority.
* I believe this is a bug in GNOME Shell/St. It is also not a recent regression, this behavior appears to have existed for a very long time.
* Because of this, styles from other extensions can override BMS's transparency rules.
* `!important` does not help here, because the imported BMS stylesheet itself has a lower priority.
* The blur is still created correctly, but it remains hidden behind the surface of the other extension.

## The Fix

* Combine all CSS files into one final stylesheet during the build instead of using runtime CSS `@import`.
* This makes all BMS styles part of the "directly loaded" extension stylesheet and gives them the expected priority.
* No runtime code/CSS needs to change. Ordering is preserved.
* This only restores the intended behavior.
* BMS still only styles common Shell surfaces after they have been selected as BMS blur targets.
* Custom inner elements remain under the control of the extension that owns them unless they use common Shell styles already supported by BMS, so nothing breaks.

## Upstream GNOME/St Bug

* Imported stylesheets should inherit the same priority as the stylesheet that imports them.
* In St that currently does not happen.
* So I believe this is an upstream St bug rather than something specific to BMS.
* Fixing it properly upstream may be a bit more complicated than it seems.
* Since the BMS workaround is only a small build-time change and also removes the runtime `@import` dependency, it makes sense to handle it here independently.

## Reproduction

The dialog created by this test extension should be blurred, but its own opaque background currently overrides BMS's transparency rule:

[stylesheet-priority-reproducer@local.shell-extension.zip](https://github.com/user-attachments/files/32163918/stylesheet-priority-reproducer%40local.shell-extension.zip)


krissedout (approved):
Thanks, this looks good!

PR #1000, opened 2026-09-15
Copyous support for popup blur

MeguMario:
This PR add support for popup blur to blur Copyous (`clipboard-dialog`), basically a version of #991 

image


tobih7:
That looks great!
I also opened [this PR](https://github.com/boerdereinar/copyous/pull/164) on the copyous side. But your solution does look a lot better visually. I’ll check if these two PRs interfere with one another later.
But I don’t think it’ll get merged anytime soon anyway, since there hasn’t been any activity in the repo for 4 months.

MeguMario:
@tobih7 I was thinking about applying the same method of adding our own style class (akin to OSK support) to workaround the styling issue, but #996 already fix that so all thanks to you too! 

I don't think it will interfere considering it's just moving support from one style class to another style class (which is also currently support by BMS).

tobih7:
You're right, the two PRs don't interfere.

src/styles/popup/base.css
```diff
@@ -76,20 +80,24 @@
...
 .bms-popup-surface-light .run-dialog,
 .bms-popup-background-dark .modal-dialog,
 .bms-popup-background-dark .run-dialog,
+.bms-popup-background-dark .clipboard-dialog,
 .bms-popup-surface-dark .modal-dialog,
-.bms-popup-surface-dark .run-dialog {
-    border: 1px solid rgba(0, 0, 0, 0.05);
+.bms-popup-surface-dark .run-dialog,
+.bms-popup-surface-dark .clipboard-dialog {
+    border: 1px solid rgba(255, 255, 255, 0.05);
```
tobih7: Why did you change this from black to white?

MeguMario: For dark theme it is like that for sometime now, tho I forgot that I put those in the light theme section lol

Let me move that away real quick

krissedout:
Thanks! The actual blur integration is much simpler now, which is nice. My concern is still the Copyous-specific styling though, we’re adding 300+ lines for its buttons, search entry, items, etc., and I’d rather that stay with Copyous instead of us maintaining its theme here.

Could we keep the small surface-detection change and separate the rest?

MeguMario:
@krissedout yeah sure I can split the stylesheet changes here. Copyous do have the option to use custom theming in its option so it probably best that those should be handle by Copyous itself. Tho doing that will missing out style changing with system color scheme option.

MeguMario:
@krissedout I splited it, if anyone want the styling here's the patch for it

[1000_style.patch](https://github.com/user-attachments/files/32413968/1000_style.patch)

krissedout:
Thanks for splitting that out! One small thing: we still need the outer clipboard-dialog background rules so the blur can actually show through. Could you keep those in the transparent/light/dark background overrides, including the preserve-theme variants? The buttons, search entry and item styling can all stay in the separate patch-- that’s the part I meant to leave with Copyous.

MeguMario:
@krissedout I added background styling back. There should be at least some styling now

Preserve-theme only affect button styling so it does not matter here (as we only style the background)

PR #1001, opened 2026-09-16
Add a separate Glow setting for Liquid Glass and rework the fresnel and specular glares

treebarkbr:
Adds a Glow slider to the Liquid Glass effect, separate from the fresnel glare and the specular glint, drawn below the glare.

The glare stays a tight, defined band that thins out and fades at the end of each curve. The specular also gets a thinner glow and a wider lobe so the glint reaches further into the corner arcs.


MuzahidGithub:
yeah screenshot looks pretty good, can you make the glare on edges thin out more gradually, currently looks same width all the way to edges specially on the top one. bottom one looks fine.

MuzahidGithub:
IMG_8993

treebarkbr:
image

image

image

How about this?

MuzahidGithub:
@treebarkbr still looks same to me. The glare should start thinning and gradually disappear while coming down half the edge length. Like Exactly a mirror copy of the bottom glare, it curves upwards to the edges while fading. Bottom glare looks perfect. Top one needs to look just a mirror copy.

krissedout:
This looks nice! Just noticed we’re removing the separate specular-strength control too though. Is that intentional now that the glare works in both directions? The description makes it sound like we’re just adding another setting.

treebarkbr:
> This looks nice! Just noticed we’re removing the separate specular-strength control too though. Is that intentional now that the glare works in both directions? The description makes it sound like we’re just adding another setting.

Yeah I removed the speculars to make the glare look similar to Liquid Glass. This PR is probably gonna move slowly cause I'm on vacation rn.

Cheers

PR #1002, opened 2026-09-16
Implement Automatic style for other components & Some bug fixes

### LINKED ISSUE #1004: [Bug] Blur effect is stuck on wallpaper after unlocking screen
Installed the extension from source.

As you can see on the video, the wallpaper is normal before locking screen, then after unlocking, the blur effect is stuck with the wallpaper. Also the lockscreen didn't go blurry, so there might be some reverse logic around.

https://github.com/user-attachments/assets/ea12939d-d0a6-4c9c-a4db-91521837fe5d

Running system is Ubuntu 26.04.1 LTS Resolute Racoon.

MeguMario:
This PR introduce the Automatic option for component styling (first seen in popup blur)

Basically the function `get_background_style()` of popup blur is splited into its own helper library, which allow other components to use that as well

image

This is not pipeline swiching according to system theme btw, that will come later (hopefully not by me)

This PR also changes & fixes some of the bug below:
- ~Add `com.rastersoft.ding` to Application blur's blacklist since we only have `com.desktop.ding` somehow (that one is for GTK4 DING)~ Added in #1006 
- Fixes styling still being apply on Panel in some override mode while Disable in Overview is still active
- Fixes Desktop Icon NG 50 and below (before GTK4 port) has its background blurred (partially fixes #1004)


krissedout (changes requested):
The DING check is a bit too broad, any window with @! in its title gets excluded from blur, even if the user explicitly whitelisted it. Can we make sure it’s actually DING first?

Automatic styling also misses Dash to Panel when switching themes outside the overview, since update_visibility skips those panels.

Also noticed the panel and dock defaults changed from static to dynamic blur. Let’s keep that out of this PR; automatic styling shouldn’t change how the blur itself works.

MeguMario:
@krissedout I've splited the dynamic blur default out, though whether this deserve a seperate PR is not really clear. I think I'll leave it to you if you want to change those (I think we should because we solved the only issue that plauge dynamic blur after all)

Also DING detection no longer detect using window-name anymore, I also fixed the Panel styling as well

krissedout:
I’d keep panel and dock blur static by default for now. Dynamic makes sense when windows actually show through, especially with floating or autohiding docks, but I’d leave that opt-in rather than make everyone pay the extra rendering cost. Fixing the rendering issues makes dynamic a better option, but doesn’t necessarily make it the better default.

There are still a few things to fix here: the title-based DING check is still present, and the new Meta.StackLayer access needs to account for GNOME versions before 51. Also, hide() can now immediately show the panel blur again through the proximity logic when a window is nearby, even with Disable in Overview enabled.

tobih7:
I think we could maybe fix the DING issue by changing how applications are identified?
Right now only `get_wm_class` is used, that works for most apps.
But there also is `get_gtk_application_id`. For most non-GTK apps this will be null so it could be ignored. But for GTK3 apps it is the actual ID, whereas wm_class is just the executable name. For GTK4 both are mostly the same.

I will test this with various apps to see whether this would be better.

MeguMario:
@krissedout I removed the title-based check. Also fixed the last two points as well.

@tobih7 it's in fact how panel blur detect DING I think, via `get_gtk_application_id`, though I think keeping thing generic would be better since we only have so many case that is like DING (and it is better to keep them excluded altogether)

tobih7:
> it's in fact how panel blur detect DING

oh, you're right

so you think we should leave how applications are identified as it is?

MeguMario:
@tobih7 unless there are other generic case that like this then yes, the way it's done right now is good enough.

krissedout:
> @krissedout I removed the title-based check. Also fixed the last two points as well.

Thanks, those fixes are in now! Two things left: update_visibility() still hides the panel background through the overview pseudo-class branch even when Disable in Overview is off. Could you make that respect the setting?

Also, I missed this earlier: the bottom stacking layer isn’t exclusive to desktop windows, so ordinary windows set to stay below use it too. Could you remove BOTTOM and its numeric fallback from the desktop check, keeping the actual desktop/DING checks?

MeguMario:
@krissedout I pushed another update, can you check if its ok now?

krissedout:
Looks good now!

PR #1003, opened 2026-09-16
Style tweaks

MeguMario:
This introduce some styling tweaks to the entire extension:

- Fixed the issue where the arrow in quick settings when the option is enabled is black in light mode
- "Clear all" button in Notification section now has a insensitive state styling
- Adjust coloring of some components to match that of Popup blur
- Adjust light mode styling
- Restore accent color highlighting on overview's search bar
- Move some dash styling in `overview.css` to `dash.css` since we also handle native dash in dock blurring now


tobih7:
With default settings, some things look odd after this PR.
- default settings, nothing changed
- system dark/light mode does not matter

## Overview
https://github.com/user-attachments/assets/82ea6243-2c8a-42c1-a2ef-c57b400013a0

- Search bar is white when empty, darker with text.
- "Searching" is low contrast.

## Application Folders
image

The styles for overview, also override font color in application folders.
When app folder style does not match general overview style (I mean light, dark, transparent), the font color is wrong and therefore hard to read.

I guess #1002 will fix this partially, because then Automatic will be used, but still the own styles for app folder should change the colors as well.

MeguMario:
@tobih7 well originally, this PR was part of #1002, so this is kinda made to compliment that, though for some of the thing you point out, it was to complement the shell light theme so it will look off when using with dark theme), and for element like the spinner, something like auto pipeline switching will be needed since right now the pipeline is fixed between light and dark mode

The appfolder one though will be fixed soon

PR #1006, opened 2026-09-19
Add com.rastersoft.ding to default blacklist for application blur

tobih7:


MeguMario:
I already added it in #1002, though I can just remove the one on my side.

MeguMario (approved):

PR #1007, opened 2026-09-19
Overview and application folder style improvements

tobih7:
## Overview

- Fixed popup styles overriding the search bar’s background and text colors when focused. Previously focusing changed the color.
- Fixed “Searching” and “No results” being forced to use dark text.
- Made navigation arrows easier to see in Light mode.
- Also see https://github.com/aunetx/blur-my-shell/pull/1003#issuecomment-5742340329

## Application folders

Fixed overview styles overriding parts of app folders. When overview and folder styles were different, folder text and edit buttons may have the wrong colors.

## Question about Transparent mode

On master opening an app folder blurred the whole screen, and Transparent removed the folder background, border, and shadow. Now the blur has the default shape again, just like Light and Dark.

Was that intentional, or did the old behavior get lost during the refactor?
If this was intentional, is Transparent even needed?
I liked the old look. If this was not intended I would be up to fix this.


MeguMario:
@tobih7 

I was fixing this and then notice this PR, I guess you save me some effort (lol).

Anyway here are some thing that I encountered

- Search bar lost its stock styling of being highlighted (it's background become brighter) when it is active
  - Current

  https://github.com/user-attachments/assets/e1cdbcbc-99a1-4883-bfa3-8c96c20d7cb2

  - No style
  
  https://github.com/user-attachments/assets/ffc64f8d-2b6d-4359-8df3-e52f1c090d50

- The spinner and text is now mismatched in light mode 
  - Though I guess for now it can stay like that until we have auto pipeline switching
  
  image

Aside from that I think this PR is good to go

For the question, well the previous PR focus on making Dark mode and Light mode experience a bit better by having something that match each style, so Transparent mode was a bit left out there (which I said I was going to fix), but it seems like you beat me up it haha. 

For the blur having its default shape, you may want to ask @krissedout for that. But in my opinion, the old behaviour was because of `Shell.BlurEffect` cannot do corner (which led to a lot of corner issue in the past), so one way to blur appfolder is to blur the whole screen behind the folder. Now with the corner out of the way, we can actually apply the blur effect on the folder background itself, which sound intentional to me

MeguMario:
@tobih7 #1008 was recently merged so you may want to resolve the conflict first, thanks!

tobih7:
@MeguMario I updated the PR. Does it look good now?
#1008 Is missing hover styling for notification group buttons for dark style. I also fixed that here.

krissedout (approved):
Looks good!

